PR #113, opened 2022-11-04
Generalise AAVE open strategy in library

zerotucks:
**Breaking changes & Post merge tasks**
* [Breaking change] Deploy updated and new actions
* Publish new library 
* All strategies return CUSTOM_OPERATION for the operation name - this is because an additional conversation is needed to address the fact generic operations that can have steps configured conditionally don't play well with fixed operation names with a fixed set of Actions.

**Changes**
* Update library interfaces
* Ensure Position class and strategies can handle varying precision levels - see PositionBalance and new helpers on Position class for handling precision
* General open strategy and underlying operation
* Refactor Open tests to use shared setup (less boiler plate)
* Add hardhat tracer for test debugging

**Ongoing issues with tests**
* Sometimes Open tests - on my machine at least - will fail. However, if run with the --trace flag using hardhat tracer to debug then pass everytime. I feel like there's some issue with hardhat's test runner. Race condition perhaps. Means the tests are not suitable for Ci for now.
* AdjustStEth tests are still failing but the logic is sound - when run with hardhat-tracer (debug mode) the tests pass every time. When run normally they don't so there's some issue with test setup most likely. Will address as part of generalising/refactoring work re adjust strategy

**Next steps**
* Refactor & generalise close strategy
* Refactor & generalise adjust strategy


contracts/actions/common/PullToProxy.sol
```diff
@@ -0,0 +1,26 @@
+pragma solidity ^0.8.15;
```
peculiarity: I still dopn't think we need this action. Whole new action only to avoid populating  a property in the library.  I'm strongly against this action. The whole idea of the `PullToken` and the data struct taht it accepts was to make it generalized.

helpers/deploy.ts
```diff
@@ -15,8 +15,9 @@ export async function createDeploy(
 
   return async (contractName: string, params: string[] = []): Promise<[Contract, string]> => {
     const contractFactory = await ethers.getContractFactory(contractName, config.signer)
-    const instance = await contractFactory.deploy(...params)
+    const instance = await contractFactory.connect(config.signer).deploy(...params)
```
peculiarity: Um you don't have to do that because ` config.signer` has been already provided  in `getContractFactory`.

helpers/init.ts
```diff
@@ -3,12 +3,24 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types'
...
 import { RuntimeConfig } from './types/common'
 
-export default async function init(hre?: HardhatRuntimeEnvironment): Promise<RuntimeConfig> {
-  const provider = hre ? hre.ethers.provider : (await import('hardhat')).ethers.provider
-  // const provider = new ethers.providers.JsonRpcProvider('http://localhost:8545', 2137)
-  const signer = provider.getSigner(0)
+// https://etherscan.io/address/0xe3dd3914ab28bb552d41b8dfe607355de4c37a51
+const accountToImpersonate = '0xe3dd3914ab28bb552d41b8dfe607355de4c37a51'
+
+export async function impersonateRichAccount(provider: providers.JsonRpcProvider) {
```
peculiarity: Why do we need this ? Also  you can create a more generic function `impersonateAccount`
```
export async function impersonatt(provider: providers.JsonRpcProvider, impersonatedAddress: string) {
    await provider.send('hardhat_impersonateAccount', [impersonatedAddress])
    const signer = provider.getSigner(address)
    const address = await signer.getAddress()
    return { signer, address }
}
 
const richAccount = '0xe3dd3914ab28bb552d41b8dfe607355de4c37a51'

impersonate(provider,richAccount)
```

or if the usage is needed in  multiple places then you can create 
```
export async function impersonateRichAccount(provider: Provider) {
  const richAccount = '0xe3dd3914ab28bb552d41b8dfe607355de4c37a51'
  return await impersonate(provider, richAccount)
}
```

That way in the future if for some reason someone want to impersonate an account there will be already a generic fgunction

helpers/init.ts
```diff
@@ -3,12 +3,24 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types'
...
   const address = await signer.getAddress()
 
+  return { signer, address }
+}
+
+export default async function init(hre?: HardhatRuntimeEnvironment): Promise<RuntimeConfig> {
+  const ethers = hre ? hre.ethers : (await import('hardhat')).ethers
+  const provider = ethers.provider
+
+  const { signer, address } = await impersonateRichAccount(provider)
```
peculiarity: Why do we need this in the `init` ? What was wrong with the current flow ?

helpers/swap/OneInchCallMock.ts
```diff
@@ -1,17 +1,38 @@
...
+    marketPrice: BigNumber = new BigNumber(testMarketPrice),
+    precision: { from: number; to: number } = { from: 18, to: 18 },
+    debug = false,
+  ) =>
   async (from: string, to: string, amount: BigNumber, slippage: BigNumber) => {
+    // EG FROM WBTC 8 to USDC 6
+    // Convert WBTC toWei at 18
+    // Apply market price
+    // Convert result back to USDC at 6
+    const precisionAdjustedToAmount = amountToWei(
```
peculiarity, line 17: That looks like complication to me.

`EG: 0.01 BTC  at 16000 USD price`
`0.01*16000 = 160`
`fromToken : amountToWei(0.01,6)`
`toToken: amountToWei(160,8)`

contracts/actions/aave/Deposit.sol
```diff
@@ -26,17 +28,25 @@ contract AaveDeposit is Executable, UseStore {
...
     DepositData memory deposit = parseInputs(data);
 
-    deposit.amount = store().readUint(bytes32(deposit.amount), paramsMap[1], address(this));
+    uint256 mappedDepositAmount = store().readUint(
+      bytes32(deposit.amount),
+      paramsMap[1],
+      address(this)
+    );
+
+    uint256 actualDepositAmount = deposit.sumAmounts
```
peculiarity, line 37: What's the idea behind these `sumAmounts` ? I guess I'm missing something here.

test/utils.ts
```diff
@@ -39,3 +39,5 @@ export async function expectRevert(expression: RegExp, tx: Promise<unknown>) {
     expect('Tx to fail', 'Tx should revert').to.be.eq('Tx succeeded')
   }
 }
+
+export const MULTIPLE_TESTING_OFFSET = 0.1
```
peculiarity: I can't see this being used anywhere...

test/deploySystem.ts
```diff
@@ -54,20 +55,24 @@ export async function deploySystem(config: RuntimeConfig, debug = false, useFall
...
 
   const [swap, swapAddress] = await deploy(CONTRACT_NAMES.common.SWAP, [
     address,
     ADDRESSES.main.feeRecipient,
     0,
     serviceRegistryAddress,
   ])
-  await swap.addFeeTier(20)
+
+  await swap.connect(signer).addFeeTier(20)
```
peculiarity: Why do you explicitly connect the signer?

test/aave/CloseStEth.test.ts
```diff
@@ -384,25 +384,31 @@ describe(`Strategy | AAVE | Close Position`, async () => {
...
           {
             amount: new BigNumber(beforeCloseUserStEthReserveData.currentATokenBalance.toString()),
+            symbol: tokens.STETH,
           },
           aaveStEthPriceInEth,
           openStrategy.simulation.position.category,
         )
 
-        closeStrategy = await strategies.aave.closeStEth(
+        closePositionMutation = await strategies.aave.close(
```
peculiarity: What's with the naming `closePositionmutation`? Can't we just name it `closePosition` and that's it ?

tasks/closePosition/index.ts
```diff
@@ -144,17 +153,19 @@ task('closePosition', 'Close stETH position on AAVE')
       },
     )
 
-    const strategyReturn = await strategies.aave.closeStEth(
+    const positionMutation = await strategies.aave.close(
```
peculiarity, line 140: Some tests is called `positionMutation` and in some `closePositionMutation`. Let's stick with one approach. Again IMO it should be just `closePosition`.

tasks/closePosition/index.ts
```diff
@@ -78,6 +79,8 @@ task('closePosition', 'Close stETH position on AAVE')
       ETH: ADDRESSES.main.ETH,
       WETH: ADDRESSES.main.WETH,
       stETH: ADDRESSES.main.stETH,
+      wBTC: ADDRESSES.main.WBTC,
```
peculiarity: I guess you can get the `mainnetAddress` from the `test/addresses`

peculiarity (commented):
Left some comments

helpers/deploy.ts
zerotucks: Have left a comment in the code.

```suggestion
    const contractFactory = await ethers.getContractFactory(contractName, {
      signer: await ethers.getSigner(await config.signer.getAddress()),
    })
```

Basically, the signer is not being persisted through when the contractFactory.deploy function is called - when using non-default signer

helpers/init.ts
BackToTheCode: We don't need it. Think I just did it in a hurry. Have removed

test/deploySystem.ts
zerotucks: Because, otherwise Hardhat is defaulting back to the default signer after resetting nodes. See -> helpers/restoreSnapshot.ts

After a node reset, we re-impersonate the rich account and have to connect the signer again.

test/aave/CloseStEth.test.ts
zerotucks: After chat swapping `Mutation` for `Transition` in tests and interfaces.

tasks/closePosition/index.ts
zerotucks: Updated

helpers/swap/OneInchCallMock.ts
zerotucks: Updated.

helpers/init.ts
zerotucks: I've updated this so you can initiailise with an impersonated account and without.

helpers/init.ts
```diff
@@ -3,12 +3,40 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types'
...
 
+  return { signer, address }
+}
+
+export async function impersonateRichAccount(provider: providers.JsonRpcProvider) {
+  // https://etherscan.io/address/0xe3dd3914ab28bb552d41b8dfe607355de4c37a51
+  const richAccount = '0xe3dd3914ab28bb552d41b8dfe607355de4c37a51'
+
+  return await impersonate(provider, richAccount)
+}
```
paszkowskiDamian: Why you need this rich account? Test account have 1000ETH isn't it enough? If you need other tokens I would use Uniswap or other exchange to get them.

zerotucks: We don't need it. You're right. I added it for convenience when I was fighting with some other issues - then it got baked in. That account owns lots of different ERC20s in quantity.

I can refactor it and replace and just use uniswap.

fenris85 (approved):

PR #119, opened 2022-11-14
Add Balancer flashloan support along with FMM

fenris85:


contracts/actions/common/TakeBalancerFlashloan.sol
```diff
@@ -0,0 +1,68 @@
...
+contract TakeBalancerFlashloan is Executable, ProxyPermission {
+  ServiceRegistry internal immutable registry;
+  address internal immutable dai;
+
+  address internal immutable balancerVault;
+
+  constructor(ServiceRegistry _registry, address _dai, address _balancerVault) {
+    registry = _registry;
+    dai = _dai;
+    balancerVault = _balancerVault;
```
peculiarity, line 28: Why don't you take the balanceVault from the Service Registry?

packages/oasis-actions/src/actions/common.ts
```diff
@@ -81,21 +81,46 @@ export function sendToken(args: { asset: string; to: string; amount: BigNumber }
...
+      actionName = CONTRACT_NAMES.common.TAKE_A_FLASHLOAN
+      flashloanCalldataType = calldataTypes.common.TakeAFlashLoan
+      flashloanCalldata = {
+        amount: args.flashloanAmount.toFixed(0),
+        dsProxyFlashloan: args.dsProxyFlashloan,
+        calls: args.calls,
+      }
+      break
+    case FLASHLOAN_TYPE.BALANCER:
+      actionName = CONTRACT_NAMES.common.TAKE_A_BALANCER_FLASHLOAN
```
peculiarity, line 106: I guess you can return the flashloan from the case block directly instead of assigniong values to some variables..

peculiarity (changes requested):

contracts/core/OperationExecutor.sol
```diff
@@ -153,4 +160,43 @@ contract OperationExecutor is IERC3156FlashBorrower {
 
     return keccak256("ERC3156FlashBorrower.onFlashLoan");
   }
+
+  function receiveFlashLoan(
```
zerotucks, line 164: We should add NatSpec descriptions

contracts/core/OperationExecutor.sol
```diff
@@ -153,4 +160,43 @@ contract OperationExecutor is IERC3156FlashBorrower {
...
+    require(msg.sender == lender, "Untrusted flashloan lender");
+
+    BalancerFlashloanData memory flData = abi.decode(data, (BalancerFlashloanData));
+    address initiator = flData.borrower;
+    IERC20 borrowedToken = tokens[0];
+
+    require(borrowedToken.balanceOf(address(this)) >= flData.amount, "Flashloan inconsistency");
+
+    if (flData.dsProxyFlashloan) {
+      borrowedToken.safeTransfer(initiator, flData.amount);
```
zerotucks, line 181: Could pull this logic out into a private method and share between flashloan cb handlers? Just because OpExecutor is probably getting fairly chunky now.

packages/oasis-actions/src/operations/aave/openStEth.ts
```diff
@@ -80,9 +82,23 @@ export async function openStEth(
     to: addresses.operationExecutor,
   })
 
+  let borrower;
+
+  switch(args.flashloanType) {
```
zerotucks, line 87: OK, I assume this PR is simply to demonstrate Balancer works. Then we can follow it up with Operation optimisations as a separate piece of work. Makes sense

peculiarity: What do you mean by that?

zerotucks: The point of Balancer is we can reduce the set of  actions and protocol touchpoints.
So instead we could have EG STETH:ETH position:
1. Flashloan WETH
2. Swap into STETH
3. Deposit STETH
4. Borrow WETH
Is handled automatically -> Repay WETH Flashloan with in-hand WETH and ETH deposited at start of t/x. We should always be able to borrow enough WETH post STETH deposit because our calculations will assume worse case scenario slippage on Swap. So, we'll be able to borrow what we need plus our initial deposit to repay FL at end.

VS Currently:
1. Flashloan DAI
2. Deposit DAI
3. Borrow WETH
4. Swap into STETH
5. Deposit STETH
6. Withdraw DAI
Repay DAI Flashloan

PR #122, opened 2022-11-16
[DO NOT MERGE] (WIP) Generalise AAVE adjust strategy in library

zerotucks:
Incomplete currently as adjust multiple down tests are failing.
Putting in review to try and allow us to unblock FE.


contracts/actions/common/uSwap.sol
```diff
@@ -211,6 +211,8 @@ contract uSwap {
 
   function swapTokens(SwapData calldata swapData) public returns (uint256) {
     IERC20(swapData.fromAsset).safeTransferFrom(msg.sender, address(this), swapData.amount);
+    uint256 balanceOnContract = IERC20(swapData.fromAsset).balanceOf(address(this));
+
```
paszkowskiDamian: Is it used?

packages/oasis-actions/src/operations/aave/close.ts
```diff
@@ -83,7 +77,7 @@ export async function closeStEth(
   })
 
   const returnFunds = actions.common.returnFunds({
-    asset: addresses.ETH,
+    asset: args.debtTokenIsEth ? ADDRESSES.main.ETH : args.debtTokenAddress,
   })
```
paszkowskiDamian: What should be passed as `debtTokenAddress` when debt token is ETH?  is it '0xeeee...'?

zerotucks: it's actually just WETH because the Operations are now written in such a way that we only have to think about WETH given if we send ETH we just wrap it right away. If we're depositing ETH we just deposit WETH. Means we don't have to mess around with AAVEWethGateway either.

contracts/actions/common/uSwap.sol
zerotucks: Nah, sorry was still working on this branch. All gone now.

PR #123, opened 2022-11-16
Versioning for Actions & Service Registry

zerotucks:
* Add versions to contract names
* Add SC entry deployment helper script
* Update local deploy system script


contracts/actions/common/uSwap.sol
```diff
@@ -211,6 +211,8 @@ contract uSwap {
 
   function swapTokens(SwapData calldata swapData) public returns (uint256) {
     IERC20(swapData.fromAsset).safeTransferFrom(msg.sender, address(this), swapData.amount);
+    uint256 balanceOnContract = IERC20(swapData.fromAsset).balanceOf(address(this));
```
peculiarity: This is not used anywhere.

contracts/core/constants/Aave.sol
```diff
@@ -3,7 +3,15 @@ pragma solidity ^0.8.15;
...
-string constant PAYBACK_ACTION = "AavePayback";
+/**
+ * @dev We do not include patch versions in contract names to allow
+ * for hotfixes of Action contracts
+ * and to limit updates to TheGraph
+ * if the types encoded in emitted events change then use a minor version and
+ * update the ServiceRegistry with a new entry
+ * and update TheGraph decoding accordingly
+ */
+string constant BORROW_ACTION = "AaveBorrow_2";
```
peculiarity, line 14: Does that mean that if there are fixes then we will add the minor version? So for now it's `AaveBorrow_2 `and if there is a fix it will be `AaveBorrow_2_1` or `AaveBorrow_2_2`?

helpers/deploy.ts
```diff
@@ -14,16 +15,20 @@ export async function createDeploy(
   const ethers = hre?.ethers || (await import('hardhat')).ethers
 
   return async (contractName: string, params: string[] = []): Promise<[Contract, string]> => {
-    const contractFactory = await ethers.getContractFactory(contractName, config.signer)
+    const contractNameWithVersionRemoved = removeVersion(contractName)
```
peculiarity, line 18: Why do we do that ?

packages/oasis-actions/src/helpers/addresses/common.json
```diff
@@ -18,6 +18,7 @@
     "feeRecipient": "0xC7b548AD9Cf38721810246C079b2d8083aba8909",
     "proxyRegistry": "0x4678f0a6958e4D2Bc4F1BAF7Bc52E8F3564f3fE4",
     "chainlinkEthUsdPriceFeed": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419",
-    "aavePriceOracle": "0xa50ba011c48153de246e5192c8f9258a2ba79ca9"
+    "aavePriceOracle": "0xa50ba011c48153de246e5192c8f9258a2ba79ca9",
+    "operationExecutor": "0x27Ec53CC96C18C777D962dDeE17b8c507238e69b"
```
peculiarity: This is the wrong address for the OperationExecutor. `0x5ab3e51608cea26090445ca89bc91628c8bb99f9` is the new one

test/aave/Open.test.ts
```diff
@@ -552,7 +552,7 @@ describe(`Strategy | AAVE | Open Position`, async function () {
     })
   })
 
-  describe('On latest block using one inch exchange and api', function () {
+  describe.skip('On latest block using one inch exchange and api', function () {
```
peculiarity: Why skipping?

PR #127, opened 2022-11-22
Reopen position in AAVE

jakubswierczek:
# [Reopen Position](https://app.shortcut.com/oazo-apps/story/6709/bug-reopen-steth-eth-doesn-t-work)

After depositing stETH we have to send action to set stETH jako collateral because: 
it's not our first deposit (After closing we have 1 wei on AAVE). 
and we can't withdraw DAI because we don't have other collateral.

### DIFF
--- a/packages/oasis-actions/src/actions/aave.ts
+++ b/packages/oasis-actions/src/actions/aave.ts
@@ -8,7 +8,7 @@ import { calldataTypes } from './types/actions'
 const createAction = ActionFactory.create
 
 export function aaveDeposit(
-  args: { amount: BigNumber | 0; asset: string },
+  args: { amount: BigNumber | 0; asset: string; setAsCollateral?: boolean },
   paramsMapping: [asset: number, amount: number] = [0, 0],
 ) {
   return createAction(
@@ -18,6 +18,7 @@ export function aaveDeposit(
       {
         asset: args.asset,
         amount: args.amount.toFixed(0),
+        setAsCollateral: args.setAsCollateral === undefined ? true : args.setAsCollateral,
       },
       paramsMapping,
     ],
--- a/tasks/createPosition/index.ts
+++ b/tasks/createPosition/index.ts
@@ -25,10 +25,6 @@ task('createPosition', 'Create stETH position on AAVE')
   .addOptionalParam('multiply', 'Required multiply', 2, types.float)
   .addFlag('dummyswap', 'Use dummy swap')
   .setAction(async (taskArgs, hre) => {
-    if (!process.env.SERVICE_REGISTRY_ADDRESS) {
-      throw new Error('SERVICE_REGISTRY_ADDRESS env variable is not set')
-    }
-
     const config = await init(hre)
 
     const serviceRegistryAddress = taskArgs.serviceRegistry || process.env.SERVICE_REGISTRY_ADDRESS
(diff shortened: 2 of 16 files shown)


tasks/createPosition/index.ts
```diff
@@ -25,10 +25,6 @@ task('createPosition', 'Create stETH position on AAVE')
   .addOptionalParam('multiply', 'Required multiply', 2, types.float)
   .addFlag('dummyswap', 'Use dummy swap')
   .setAction(async (taskArgs, hre) => {
-    if (!process.env.SERVICE_REGISTRY_ADDRESS) {
-      throw new Error('SERVICE_REGISTRY_ADDRESS env variable is not set')
```
peculiarity, line 29: Why did you remove this?

tasks/createPosition/index.ts
```diff
@@ -83,39 +79,43 @@ task('createPosition', 'Create stETH position on AAVE')
...
       config.provider,
     )
-
-    const aavePriceOracle = new hre.ethers.Contract(
-      ADDRESSES.main.aavePriceOracle,
-      aavePriceOracleABI,
-      config.provider,
-    )
-
+    new hre.ethers.Contract(ADDRESSES.main.aavePriceOracle, aavePriceOracleABI, config.provider)
```
peculiarity: Either I'm blind or this does absolutely nothing. I mean it creates an instance and destroys it because nothing uses it.

tasks/createPosition/index.ts
```diff
@@ -83,39 +79,43 @@ task('createPosition', 'Create stETH position on AAVE')
       AAVELendigPoolABI,
       config.provider,
     )
+
+    config.provider.on(aaveLendingPool.filters.Deposit(ADDRESSES.main.DAI), (log, event) => {
```
peculiarity: Why do you need ths?

test/aave/ReopenStEth.test.ts
```diff
@@ -0,0 +1,318 @@
...
+  })
+
+  describe('Should close position with real oneInch', () => {
+    const slippage = new BigNumber(0.1)
+
+    let openTxStatus: boolean
+    let closeTxStatus: boolean
+    let reopenTxStatus: boolean
+
+    before(async function () {
```
peculiarity, line 241: What's the point of having before when you can skip before and have on `it` test scenario with 3 expectations in it ?

tasks/createPosition/index.ts
jakubswierczek: We have it in parameteres: 
`  .addOptionalParam<string>('serviceRegistry', 'Service Registry address')`
So we don't need to have it in env

tasks/createPosition/index.ts
jakubswierczek: to delete

test/aave/ReopenStEth.test.ts
jakubswierczek: It's tricky. `this.skip()` in that example is `describe` so we are skipping all tests in that describe. 
If we will use `() => ` instead of `function() {}` that would not be working.

tasks/createPosition/index.ts
peculiarity: Well I understand but as you can see the param is optional so I might not pass it and since we won't have a place to read the ServiceRegistry from  the ask will fail. The param is created ( my assumption ) just to override the `.env` configuration if the developer needs that for some reason.
Also even if you rename `addOptionalParam` to just `addPAram` that way every single time i have to go around and look for the serviceRegistry address somewhere.

test/aave/ReopenStEth.test.ts
peculiarity: Lol. I wasn't refering to `.skip`. 
```
describe((`Strategy | AAVE `) => {
   it('should reopen a position',m async () => {
   // open positon
    expect(openTxStatus).to.be.true
    // close positon
    expect(closeTxStatus).to.be.true
    // open position
    expect(reopenTxStatus).to.be.true
   })
})
```

peculiarity: Also I wouldn't rely on `txStatus` but rather values in the position. It's possible to call a non existing contract or contract with some fallback function implemented and proxy will return successful call but you would have done absolutely nothing. 
I can remember multiple scenarios like this in our application on production

tasks/createPosition/index.ts
jakubswierczek: I will move it after reading it from env or params

test/aave/ReopenStEth.test.ts
jakubswierczek: 👍

jakubswierczek: I will change the tests

packages/oasis-actions/src/actions/aave.ts
```diff
@@ -18,6 +18,7 @@ export function aaveDeposit(
       {
         asset: args.asset,
         amount: args.amount.toFixed(0),
+        setAsCollateral: args.setAsCollateral === undefined ? true : args.setAsCollateral,
```
fenris85, line 21: I'm thinking if this should be true as default.. This is our main use case at the moment, right?

fenris85 (approved):

PR #130, opened 2022-11-25
Optional actions in operation

fenris85:
Optional actions in operation are marked by passing additional array when creating operation in the OperationRegistry, e.g.:

`[ Action1, Action2, Action3, Actions4]` - actions hashes list
`[false, true, false, false]` - optionality list

In above case `Action2` can be skipped. 

Skipping when forming an operation on FE can be achieved by adding param `skipped = true`. This action still has to exist in operation, but all params can be omitted. It requires only actionHash and `skipped` param.

### DIFF
--- a/contracts/core/OperationsRegistry.sol
+++ b/contracts/core/OperationsRegistry.sol
@@ -43,13 +43,11 @@ contract OperationsRegistry {
 
   /**
    * @notice Adds an Operation's Actions keyed to a an operation name
-   * @param name The Operation name
-   * @param actions An array the Actions the Operation consists of
-   * @param optional An array indicating given Action can be skipped
+   * @param operation Struct with Operation name, actions and their optionality
    */
-  function addOperation(string memory name, bytes32[] memory actions, bool[] memory optional) external onlyOwner {
-    operations[name] = StoredOperation(actions, optional, name);
-    emit OperationAdded(name);
+  function addOperation(StoredOperation calldata operation) external onlyOwner {
+    operations[operation.name] = operation;
+    emit OperationAdded(operation.name);
   }
 
   /**
--- a/contracts/core/OperationsRegistry.sol
+++ b/contracts/core/OperationsRegistry.sol
@@ -5,12 +5,14 @@ import { OPERATIONS_REGISTRY } from "./constants/Common.sol";
 
 struct StoredOperation {
   bytes32[] actions;
+  bool[] optional;
   string name;
 }
 
 /**
  * @title Operation Registry
- * @notice Stores the Actions that constitute a given Operation
+ * @notice Stores the Actions that constitute a given Operation and information if an Action can be skipped
+
  */
 contract OperationsRegistry {
   mapping(string => StoredOperation) private operations;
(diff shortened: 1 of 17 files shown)


contracts/core/OperationsRegistry.sol
```diff
@@ -5,12 +5,14 @@ import { OPERATIONS_REGISTRY } from "./constants/Common.sol";
 
 struct StoredOperation {
   bytes32[] actions;
+  bool[] optional;
   string name;
 }
```
peculiarity, line 12:
```
struct Action {
  hash: bytes32,
  isSKippable: bool,
 }
 ```

Comment (peculiarity) [reply contracts/core/OperationsRegistry.sol:12]:
```
struct Operation {
   actions: Action[],
   string: name
}
```

File: contracts/core/OperationsRegistry.sol [hunk 2]
```diff
@@ -43,21 +45,24 @@ contract OperationsRegistry {
    * @notice Adds an Operation's Actions keyed to a an operation name
    * @param name The Operation name
    * @param actions An array the Actions the Operation consists of
+   * @param optional An array indicating given Action can be skipped
    */
-  function addOperation(string memory name, bytes32[] memory actions) external onlyOwner {
-    operations[name] = StoredOperation(actions, name);
+  function addOperation(string memory name, bytes32[] memory actions, bool[] memory optional) external onlyOwner {
```
peculiarity: ```addOperation(Operation calldata operation) external onlyOwner```

test/common/OptionalActions.test.ts
```diff
@@ -0,0 +1,163 @@
...
+import { executeThroughProxy } from '../../helpers/deploy'
+import { restoreSnapshot } from '../../helpers/restoreSnapshot'
+import { ServiceRegistry } from '../../helpers/serviceRegistry'
+import { RuntimeConfig } from '../../helpers/types/common'
+import { OperationsRegistry } from '../../helpers/wrappers/operationsRegistry'
+import { testBlockNumber } from '../config'
+import { DeployedSystemInfo } from '../deploySystem'
+import { initialiseConfig } from '../fixtures/setup'
+
+const createAction = ActionFactory.create
```
peculiarity, line 19: IMO tests are not written well. 
Let's for example comment out the call

```
// if(!calls[current].skipped) {
//   address target = registry.getServiceAddress(calls[current].targetHash);
//   target.execute(calls[current].callData);
// }
 ```
 Which is in the `OperationExecutor`. These test will pass. 
 I'm not sure what exactly is being test. Also you shouldn't be testing a given contract through proxy calls ( I mean one contract, calling the other contract ) and verifying if the TX is successful.

peculiarity: Write tests that given `addOperation` and some params ` `getOperation` will return same values.
That will be under `OperationsRegistry.test.ts`
We should have `OperationsExecutor` which must contain logic about testing the login within `executeOp`. 
You can call an action for example that will fire an event and check if the TX contains that event or if state is changed or smth like that. 
We have only integration tests which is not good.

peculiarity (changes requested):

contracts/core/OperationsRegistry.sol
zerotucks: Kuba mentioned he's already looked into this. And dynamic arrays of structs create issues elsewhere

test/common/OptionalActions.test.ts
zerotucks: Broadly agree. I'm going to ask Frank if we can have a ticket for this next sprint.

zerotucks (approved):

PR #131, opened 2022-11-28
Added a script to quickly update the block number

marcinciarka:
`yarn update-block`

1. Creates a backup
2. Updates a block number (-50 for _reasons_)

### DIFF
--- a/update-block-number.ts
+++ b/update-block-number.ts
@@ -1,0 +1,43 @@
+const fetch = require('node-fetch')
+const fs = require('fs')
+const path = require('path')
+const mainnetInfoUrl = 'https://blockexplorer.one/ajax/eth/mainnet/info'
+const envFile = `${path.join(__dirname)}/.env`
+const envFileBackup = `${path.join(__dirname)}/.env_bak`
+const newLine = '\n'
+
+const updateNumber = () => {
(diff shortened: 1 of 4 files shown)


update-block-number.ts
```diff
@@ -0,0 +1,43 @@
...
+      const blockHeight = res.height
+      console.log('😊 Found block height:', blockHeight)
+      const envFileContents = fs.readFileSync(envFile, 'utf8')
+      const envFileContentLines = envFileContents.split(newLine)
+      const filteredEnvFileContentLines = envFileContentLines.map(line =>
+        line.startsWith('BLOCK_NUMBER') ? `BLOCK_NUMBER=${blockHeight - 50}` : line,
+      )
+      try {
+        fs.writeFileSync(envFile, filteredEnvFileContentLines.join(newLine))
+        console.log('🙌 Updated the block number with', blockHeight - 50, '(latest block minus 50)')
```
sennett: This is awesome.  Could go for -10 here though?  That is early enough to avoid reorgs and the liquidity on the swapping protocols is less likely to drift.

sennett (approved):

update-block-number.ts
marcinciarka: Done

PR #134, opened 2022-11-29
feat: Allow user to deposit into an existing position or new  one. Sw…

peculiarity:
…ap is optional.


packages/oasis-actions/src/helpers/constants.ts
```diff
@@ -5,6 +5,7 @@ export const TYPICAL_PRECISION = 18
 export const ZERO = new BigNumber(0)
 export const ONE = new BigNumber(1)
 export const TEN = new BigNumber(10)
+export const HUNDRED = new BigNumber(10)
```
zerotucks:
```suggestion
export const HUNDRED = new BigNumber(100)
```

paszkowskiDamian: Is it right?

packages/oasis-actions/src/operations/aave/deposit.ts
```diff
@@ -0,0 +1,104 @@
...
+// TODO: Probably put a generic about the tokens i.e DepositArgs<Tokens> and provide AAVETokens
+// TODO: Take into consideration the precision for the tokens. So indeed might be a good idea for the Tokens generics
+interface DepositArgs {
+  // - either for a swap where the `entryToken` will be exchanged for the `depositToken`
+  // - or it will be directly deposited in the protocol
+  entryToken: string
+  // - either used for a swap if `entryToken` is swapped for `depositToken`
+  // - or it will be directly deposited in the protocol
+  amount: BigNumber
+  // - if it's omitted that means that the `entryToken` with bbe used in the deposit
```
paszkowskiDamian: `with bbe` or `will be`?

packages/oasis-actions/src/operations/aave/deposit.ts
```diff
@@ -0,0 +1,104 @@
...
+  swapArgs?: SwapArgs
+}
+export async function deposit({
+  entryToken,
+  depositToken,
+  amount,
+  depositorAddress,
+  swapArgs,
+}: DepositArgs) {
+  const isSwapNeeded = !!depositToken
```
paszkowskiDamian: What is your opinion on being more explicit, instead of not passing `depositToken` consumer should always provide both. In this case `const isSwapNeeded = entryToken !== depositToken`?

packages/oasis-actions/src/helpers/constants.ts
peculiarity: Absolutely not

packages/oasis-actions/src/operations/aave/deposit.ts
peculiarity: Was thinking about it. The idea is to not add all sort of properties when we can achieve something with the existing ones. If you think it's too confusing / complicating and etc I can agree to just provide a boolean value and that's it :) . Probably being more explicit is better and will lead to less issues / bugs in the future.

packages/oasis-actions/src/operations/aave/deposit.ts
peculiarity: Absolutely `will be`

helpers/swap/OneInchCall.ts
```diff
@@ -22,14 +22,14 @@ export const getOneInchCall =
...
-      console.log('slippage %', slippageAsPercentage.toString())
+      console.log(`slippage ${slippageAsPercentage.toString()}%`)
     }
 
     return {
       toTokenAddress: to,
       fromTokenAddress: from,
       minToTokenAmount: new BigNumber(response.toTokenAmount)
-        .times(one.minus(slippage))
+        .times(one.minus(slippage.div(new BigNumber(100))))
```
zerotucks, line 34: I think if we change this line then we need to change line 25 and update all the tests to so that the passed slippages are much larger.

For example, tests are often passed 0.1 (which is currently equivalent to 10%) but if we change this line it'll be equivalent 0.1%

zerotucks: It probably makes sense to align with 1inch api

packages/oasis-actions/src/helpers/calculations/Position.ts
```diff
@@ -49,8 +50,8 @@ export type Swap = {
   minToTokenAmount: BigNumber
   tokenFee: BigNumber
   collectFeeFrom: 'sourceToken' | 'targetToken'
-  sourceToken: { symbol: string; precision: number }
-  targetToken: { symbol: string; precision: number }
+  sourceToken: TokenDef<AAVETokens>
```
zerotucks, line 53: Nice

zerotucks (approved):

tenderly.yaml
```diff
@@ -17,4 +17,4 @@ exports:
       istanbul_block: 0
       berlin_block: 0
       london_block: 0
-project_slug: ''
+project_slug: 'georgi3/compound-integration'
```
zerotucks: Not sure if you want to commit that

packages/oasis-actions/src/operations/aave/deposit.ts
paszkowskiDamian: I didn't mean that we should provide a `isSwapNeeded` flag. I just meant that `depositToken` should always be provided. Imagine invoking it on frontend. 

```
deposit({
   ....
   entryToken, 
   depositToken: depositToken === entryToken ? undefined : depositToken
})

// where it could be like 

deposit({
    entryToken,
   depositToken, 
})

```

paszkowskiDamian: Another case what if they are both defined but are different, the logic would assume the swap is not necessary thus the tx would fail or the outcome will be unpredictable.

peculiarity: On the frontend you don't call `deposit` from `operation` but `deposit` from `strategy`.

peculiarity: > Another case what if they are both defined but are different, the logic would assume the swap is not necessary thus the tx would fail or the outcome will be unpredictable.

I don't follow. There are even tests exploring different scenarios where if the two are different there is a swap.

peculiarity:
@paszkowskiDamian  Can you please verify the simulation part?

peculiarity:
@paszkowskiDamian works on a separate branch which includes these changes

PR #136, opened 2022-11-30
Integrate optional actions w/ library

zerotucks:
NOTE: Breaks create/close position tasks as they fetch the deployed version of the Operation Executor from mainnet


package.json
```diff
@@ -65,7 +65,7 @@
     "ethereum-waffle": "^3.1.1",
     "ethers": "^5.0.19",
     "googleapis": "107",
-    "hardhat": "2.9.5",
+    "hardhat": "2.8.0",
```
fenris85, line 68: why downgrading hardhat?

packages/oasis-actions/src/operations/aave/close.ts
```diff
@@ -93,9 +100,12 @@ export async function close(
...
       withdrawDAIFromAAVE,
       unwrapEth,
-      returnFunds,
+      returnDebtFunds,
+      returnCollateralFunds,
     ],
   })
 
-  return { calls: [takeAFlashLoan], operationName: 'CUSTOM_OPERATION' }
+  const operationName = OPERATION_NAMES.aave.CLOSE_POSITION
```
fenris85: Is assigning to a new variable needed here? If so, then you can do `return { operationName, calls: [takeAFlashLoan] }`

packages/oasis-actions/src/operations/aave/decreaseMultiple.ts
```diff
@@ -74,21 +76,25 @@ export async function decreaseMultiple(
...
-      withdrawDAIFromAAVE,
-    ],
+    calls: flashloanCalls,
   })
 
-  return { calls: [takeAFlashLoan], operationName: 'CUSTOM_OPERATION' }
+  const calls = [takeAFlashLoan]
+  const operationName = OPERATION_NAMES.aave.DECREASE_POSITION
+
+  return { calls, operationName }
```
fenris85: Here it's better ;),  but I would still combine it to single return

packages/oasis-actions/src/operations/aave/increaseMultiple.ts
```diff
@@ -127,11 +127,10 @@ export async function increaseMultiple(
   })
 
   const calls = [takeAFlashLoan]
-  use.pullDebtTokensInToProxy && calls.unshift(pullDebtTokensToProxy)
-  use.pullCollateralInToProxy && calls.unshift(pullCollateralTokensToProxy)
+  const operationName = OPERATION_NAMES.aave.INCREASE_POSITION
```
fenris85: same as above

package.json
zerotucks: tenderly export support. 2.9.5 doesn't work with tenderly. And the only think 2.8 breaks is Maker tests which don't seem relevant right now.

So, perhaps we leave at 2.8.0 so people don't have to keep downgrading during development? And then address Maker tests once they become relevant again?

jakubswierczek (approved):

PR #143, opened 2022-12-09
Deposit and borrow

paszkowskiDamian:


helpers/swap/OneInchCall.ts
```diff
@@ -22,14 +22,14 @@ export const getOneInchCall =
...
-      console.log('slippage %', slippageAsPercentage.toString())
+      console.log(`slippage ${slippageAsPercentage.toString()}%`)
     }
 
     return {
       toTokenAddress: to,
       fromTokenAddress: from,
       minToTokenAmount: new BigNumber(response.toTokenAmount)
-        .times(one.minus(slippage))
+        .times(one.minus(slippage.div(new BigNumber(100))))
```
halaprix, line 32: does it need to be divided ? it's `slippage` not `slippageAsPercentage`

PR #145, opened 2022-12-11
Fix Adjust, Add new 1inch tests

zerotucks:


test/aave/Adjust.test.ts
```diff
@@ -538,13 +538,13 @@ describe(`Strategy | AAVE | Adjust Position`, async function () {
       })
 
       it('Should collect fee', async () => {
-        const feeRecipientUSDCBalanceAfter = await balanceOf(
+        const feeRecipientWETHBalanceAfter = await balanceOf(
```
sennett, line 537: It's checking USDC balance here I think?

packages/oasis-actions/src/strategies/aave/adjust.ts
```diff
@@ -110,7 +128,7 @@ export async function adjust(
         oazo: new BigNumber(FEE),
       },
       prices: {
-        market: quoteMarketPrice,
+        market: isIncreasingRisk ? quoteMarketPrice : ONE.div(quoteMarketPrice),
```
paszkowskiDamian: I'm not sure that inverse of the quote market price is right. Image eth -> steth, you can swap 1:1 eth to stEth but stEth -> eth is usually less then 1.

packages/oasis-actions/src/strategies/aave/adjust.ts
```diff
@@ -36,10 +36,22 @@ export async function adjust(
...
+    )
+  ) {
+    isIncreasingRisk = false
+  }
+
+  const estimatedSwapAmount = amountToWei(
+    new BigNumber(1),
+    isIncreasingRisk ? args.debtToken.precision : args.collateralToken.precision,
+  )
+
```
paszkowskiDamian: Maybe instead of checking `isIncreasingRisk` there should be just one please where you assign `fromToken` and `toToken`

packages/oasis-actions/src/strategies/aave/adjust.ts
```diff
@@ -411,21 +422,16 @@ async function _decreaseRisk({
...
+  ).div(actualMarketPriceWithSlippage)
+  const debtTokenAmountInBaseUnits = amountToWei(debtTokenAmount, args.debtToken.precision)
 
   const newDebt = {
-    amount: existingPosition.debt.amount.minus(debtTokenAmountAfterSwapInWei),
+    amount: BigNumber.max(existingPosition.debt.amount.minus(debtTokenAmountInBaseUnits), ZERO),
     precision: existingPosition.debt.precision,
-    symbol: target.position.collateral.symbol,
+    symbol: target.position.debt.symbol,
   }
```
paszkowskiDamian: I had the same issues/thoughts in withdraw. What should happen when user wants to withdraw more than they have? Throw error? just return zero? 🤷

test/aave/Close.test.ts
```diff
@@ -822,4 +822,194 @@ describe(`Strategy | AAVE | Close Position`, async () => {
...
+        const openPositionTransition = await strategies.aave.open(
+          {
+            depositedByUser: {
+              collateralToken: { amountInBaseUnit: depositWBTCAmount },
+            },
+            positionArgs: {
+              positionId: 123,
+              positionType: 'Multiply',
+              protocol: 'AAVE' as const,
+            },
```
paszkowskiDamian: we discussed `positionId` already.

But I think `positionType` nad `protocol` can be hardcoded within strategy. The consumer of the library shouldn't be able to set it to anything else while using `aave.open`

packages/oasis-actions/src/strategies/aave/adjust.ts
zerotucks: Have removed most of the flags and just working with from and to

packages/oasis-actions/src/strategies/aave/adjust.ts
zerotucks: Yea, so occasionally when decreasing we're returning more from the swap than the remaining debt balance. The swap simulation (max slippage) is always pessimistic so when the swap out performs it we have to handle this yea. Sometimes no debt left.

test/aave/Close.test.ts
zerotucks: Yes, positionType is the only big Q. Do we handle within library or pass from FE

paszkowskiDamian (approved):

test/aave/Close.test.ts
sennett: I currently think the opposite from Damian here - I think we should pass from FE, that way the library is as generic as possible and this same multiply mechanic can be used across both Multiply and Earn positions.  I guess the other way would be to have the same API twice, once for Earn and once for Multiply?  That seems to be more code and more copy-paste that configuring position type from the outside.

paszkowskiDamian:
https://github.com/OasisDEX/oasis-earn-sc/pull/145#discussion_r1048139058

@sennett sure I wasn't aware that open is used for both Earn and Multiply. I just wanted to point out that we cannot trust user that they'll provide valid positionId, and the protocol don't need to be passed by frontend because `aave.open` always opens AAVE position

PR #148, opened 2022-12-13
Remove positionId from event

paszkowskiDamian:
- Remove position id from event
- Use msg.sender as proxy in the event


packages/oasis-actions/src/operations/aave/open.ts
```diff
@@ -119,11 +113,12 @@ export async function open({
     to: addresses.operationExecutor,
   })
 
+  const protocol: Protocol = 'AAVE'
+  const positionType: PositionType = 'Multiply'
```
zerotucks: What if it's Earn?

zerotucks (approved):

PR #151, opened 2022-12-13
Fix tests in tests/common, skip tests in tests/maker, Refactor AAVE Open Position tests

zerotucks:
**Description**
* Fixes failing unit tests in /common
* Refactors AAVE Open test to clearly use specific scenarios. The goal here is to make our integration tests easier to reason about and maintain
* Skips Maker tests for now - which currently fail due to how we get oracle prices for these tests - current version of Hardhat doesn't work with this approach. Updating hardhat also breaks tenderly. So, pushing pause on these tests for now until they're more relevant.


test/aave/Open.test.ts
```diff
@@ -34,266 +31,356 @@ import { amountToWei, balanceOf } from '../../helpers/utils'
...
+      config,
+      provider,
+      blockNumber: testBlockNumber,
+      useFallbackSwap,
+    })
+    const system = snapshot.deployed.system
+
+    /**
+     * Test setup
+     * Swap test account for ETH for tokens required for transaction
```
zerotucks:
```suggestion
     * Swap test account ETH for tokens required for transaction
```

helpers/utils.ts
```diff
@@ -182,3 +183,22 @@ export function asPercentageValue(value: BigNumber.Value, base: BigNumber.Value)
...
+}
+
+export async function getRecentBlock({
+  provider,
+  offset,
+  roundToNearest,
+}: {
+  provider: JsonRpcProvider
+  offset: number
+  roundToNearest?: number
```
jakubswierczek, line 198: I think you can add a default value in parameter definitions.

packages/oasis-actions/src/operations/aave/borrow.ts
```diff
@@ -3,7 +3,7 @@ import BigNumber from 'bignumber.js'
 import * as actions from '../../actions'
 import { ADDRESSES } from '../../helpers/addresses'
 import { OPERATION_NAMES } from '../../helpers/constants'
-import { Address } from '../../strategies/types'
+import { Address } from '../../strategies/types/StrategyParams'
```
jakubswierczek: is having longer import necessary?

jakubswierczek: I think having one place to import types is for better developer experience

packages/oasis-actions/src/strategies/aave/depositBorrow.ts
```diff
@@ -9,8 +9,9 @@ import * as operations from '../../operations'
 import { AAVEStrategyAddresses } from '../../operations/aave/addresses'
 import { BorrowArgs } from '../../operations/aave/borrow'
 import { DepositArgs } from '../../operations/aave/deposit'
-import { Address, IPositionTransition, IPositionTransitionDependencies, SwapData } from '../types'
+import { IPositionTransition, SwapData } from '../types'
 import { AAVETokens } from '../types/aave'
+import { Address, IPositionTransitionDependencies } from '../types/StrategyParams'
```
jakubswierczek: The concept of the one `index.ts` in the `types` folder is beneficial because we don't need to change all imports after changing the file name in this example.

test/aave/Close.test.ts
```diff
@@ -1,5 +1,6 @@
 import { JsonRpcProvider } from '@ethersproject/providers'
 import { TYPICAL_PRECISION } from '@oasisdex/oasis-actions'
+import { PositionType } from '@oasisdex/oasis-actions/lib/src/strategies/types'
```
jakubswierczek: we should export that just from `@oasisdex/oasis-actions/strategies/types` or something like that and no import form the `src` folder.

zerotucks: Funnily enough that import needs updating anyway.

Trying to update all imports to /oasis-actions/src -> so that tests and rest of repo are not dependent on the library being built first.

packages/oasis-actions/src/strategies/aave/depositBorrow.ts
zerotucks: Yea, everything is exported from an index. Just think there's a few old import lines hanging around. I probably updated that one with a find/replcae

packages/oasis-actions/src/operations/aave/borrow.ts
zerotucks: No it is not 😀 have changed.

helpers/utils.ts
zerotucks: done

jakubswierczek (approved):

PR #163, opened 2022-12-21
remove some logging, wrong amount of collateral to deposit

jakubswierczek:


packages/oasis-actions/src/helpers/calculations/Position.ts
```diff
@@ -143,6 +145,22 @@ export class Position implements IPosition {
...
+  }
+
+  public get maxCollateralToWithdraw() {
+    const maxLoanToValue = this.category.maxLoanToValue
+    const minimumCollateral = this.debt.normalisedAmount.div(
+      maxLoanToValue.times(this._oraclePriceForCollateralDebtExchangeRate),
+    )
+    return this.collateral.amount.minus(minimumCollateral)
+  }
+
```
paszkowskiDamian, line 163: In general don't you think this should be a utility function? `maxDebtToBorrow(position): BigNumber`?

paszkowskiDamian (approved):

PR #167, opened 2023-01-17
Fix: 1wei error on Close (ETH:USDC) and also in Adjust

zerotucks:
https://app.shortcut.com/oazo-apps/story/7268/investigation-small-usdc-transfers-in-multiply-txs

* Corrects rounding issue with tokenFee in Position.ts which was affecting adjust strategies
* Removes the rounding issue from the after fees swap amount for the close strategy
* Adds a 1inch test for ETH:USDC multiply pair
* Refactor Close strategy for readability.

### DIFF
--- a/packages/oasis-actions/src/helpers/calculations/Position.ts
+++ b/packages/oasis-actions/src/helpers/calculations/Position.ts
@@ -491,11 +491,11 @@ export class Position implements IPosition {
     const sourceFee = this._denormaliseAmount(
       normalisedSourceFee,
       debtTokenIsSourceToken ? this.debt.precision : this.collateral.precision,
-    )
+    ).integerValue(BigNumber.ROUND_DOWN)
     const targetFee = this._denormaliseAmount(
       normalisedTargetFee,
       debtTokenIsSourceToken ? this.collateral.precision : this.debt.precision,
-    )
+    ).integerValue(BigNumber.ROUND_DOWN)
 
     const fromTokenPrecision = isIncreasingRisk
       ? targetPosition.debt.precision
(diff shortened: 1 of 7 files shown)


packages/oasis-actions/src/helpers/calculations/Position.ts
```diff
@@ -134,6 +134,14 @@ export class Position implements IPosition {
     this.category = category
   }
 
+  /*
+   * Certain strategies don't require the adjust method but still need to calculateFees in a consistent way.
+   * This functionality is now standardised and can be accessed within strategies via a static method
+   * */
+  public static calculateFee(amount: BigNumber, fee: BigNumber, feeBase: BigNumber): BigNumber {
```
peculiarity: There is no need to define it here. There is already one defined in `helpers/index.ts`
I think it's the saem used in close or open...

packages/oasis-actions/src/helpers/calculations/Position.ts
```diff
@@ -656,7 +664,7 @@ export class Position implements IPosition {
   }
 
   private _calculateFee(amount: BigNumber, fee: BigNumber): BigNumber {
```
peculiarity: That whole method will be irrelevant and you can just delete it.

packages/oasis-actions/src/strategies/aave/close.ts
```diff
@@ -45,12 +45,17 @@ export async function close(
   const FEE_BASE = 10000
   const collectFeeFrom = args.collectSwapFeeFrom ?? 'sourceToken'
   const swapAmountBeforeFees = args.collateralAmountLockedInProtocolInWei
-  const fee = calculateFee(swapAmountBeforeFees, FEE, FEE_BASE)
 
-  const swapAmountAfterFees = swapAmountBeforeFees.minus(
-    collectFeeFrom === 'sourceToken' ? fee : ZERO,
+  const fee = Position.calculateFee(
```
peculiarity: I think the older way was better.   Also the off by wei was in adjust i think so not quire sure if this fixes the issue

peculiarity (approved):

jakubswierczek (approved):

PR #168, opened 2023-01-19
Fix: Library handles fee collection not client

zerotucks:
https://app.shortcut.com/oazo-apps/story/7574/library-collecting-fee

* Strategies now self-determine what token they'll take the fee from for a swap
* All strategies in /strategies folder updated with args removed
* New acceptedFeeToken helper returns either sourceToken or targetToken when determining fee
* Tests updated accordingly

### DIFF
--- a/packages/oasis-actions/src/config/acceptedFeeTokensConfig.ts
+++ b/packages/oasis-actions/src/config/acceptedFeeTokensConfig.ts
@@ -1,0 +1,12 @@
+/* Selection of token symbols and addresses that we're happy to accept as a fee */
+import { ADDRESSES } from '../helpers/addresses'
+
+export const acceptedTokenSymbols = ['ETH', 'WETH', 'USDT', 'USDC', 'WBTC', 'DAI']
+export const acceptedTokenAddresses = [
+  ADDRESSES.main.ETH,
+  ADDRESSES.main.WETH,
+  ADDRESSES.main.USDT,
+  ADDRESSES.main.USDC,
+  ADDRESSES.main.WBTC,
+  ADDRESSES.main.DAI,
+]
--- a/packages/oasis-actions/src/config/acceptedFeeTokensConfig.ts
+++ b/packages/oasis-actions/src/config/acceptedFeeTokensConfig.ts
@@ -1,12 +1,29 @@
 /* Selection of token symbols and addresses that we're happy to accept as a fee */
 import { ADDRESSES } from '../helpers/addresses'
 
-export const acceptedTokenSymbols = ['ETH', 'WETH', 'USDT', 'USDC', 'WBTC', 'DAI']
-export const acceptedTokenAddresses = [
-  ADDRESSES.main.ETH,
-  ADDRESSES.main.WETH,
-  ADDRESSES.main.USDT,
-  ADDRESSES.main.USDC,
-  ADDRESSES.main.WBTC,
-  ADDRESSES.main.DAI,
+export const acceptedTokens = [
+  {
+    symbol: 'USDC',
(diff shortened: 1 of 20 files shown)


packages/oasis-actions/src/helpers/acceptedFeeToken.ts
```diff
@@ -0,0 +1,28 @@
...
+
+type TokenSymbolOrAddress = string
+
+interface Props {
+  fromToken: TokenSymbolOrAddress
+  toToken: TokenSymbolOrAddress
+}
+
+/**
+ * Prefers sourceToken over targetToken
```
jakubswierczek, line 11: I think it should prefer stablecoin over ETH e.g. Let's discuss it  with Frank.

packages/oasis-actions/src/config/acceptedFeeTokensConfig.ts
```diff
@@ -0,0 +1,12 @@
+/* Selection of token symbols and addresses that we're happy to accept as a fee */
+import { ADDRESSES } from '../helpers/addresses'
+
+export const acceptedTokenSymbols = ['ETH', 'WETH', 'USDT', 'USDC', 'WBTC', 'DAI']
+export const acceptedTokenAddresses = [
```
jakubswierczek: maybe that can be a hierarchy?

packages/oasis-actions/src/helpers/acceptedFeeToken.ts
```diff
@@ -0,0 +1,28 @@
...
+export function acceptedFeeToken({ fromToken, toToken }: Props) {
+  if (acceptSourceToken(fromToken)) return 'sourceToken'
+  if (acceptTargetToken(toToken)) return 'targetToken'
+
+  const fallbackTokenType = 'sourceToken'
+  return fallbackTokenType
+}
+
+function acceptSourceToken(sourceToken: TokenSymbolOrAddress) {
+  return acceptedTokenSymbols.includes(sourceToken) || acceptedTokenAddresses.includes(sourceToken)
```
jakubswierczek: if `acceptedTokenSymbols` and `acceptedTokenAddresses` are in the correct order then we can return an index of the given token. 
then `acceptedFeeToken` needs to return the token with a smaller number.

packages/oasis-actions/src/strategies/aave/close.ts
```diff
@@ -134,8 +148,8 @@ export async function close(
       flags: flags,
       swap: {
         ...swapData,
-        tokenFee: fee,
-        collectFeeFrom: args.collectSwapFeeFrom ?? 'sourceToken',
+        tokenFee: collectFeeFrom === 'sourceToken' ? fee : postSwapFee,
```
jakubswierczek: can be just `fee.add(postSwapFee)`

jakubswierczek (commented):
Great work 👍 
Let's discuss that with correct order of tokens.

packages/oasis-actions/src/helpers/acceptedFeeToken.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * Prefers sourceToken over targetToken
+ * Accepts args as either a token symbol or in address format
+ */
+export function acceptedFeeToken({ fromToken, toToken }: Props) {
+  if (acceptSourceToken(fromToken)) return 'sourceToken'
+  if (acceptTargetToken(toToken)) return 'targetToken'
+
```
jakubswierczek, line 21: please log an error here to know that we are missing something.

jakubswierczek (approved):

PR #173, opened 2023-01-25
[SC-7451] Payback with return funds in case of something left on proxy

jakubswierczek:
## Changes 
- Fixed `MaxCollateralToWithdraw` in Position
- Added `DebtToPaybackAll` in Position
- Removed `SendToken` from `WithdrawPayback` operation (It was there only for the first implementation of DPM)
- Added `UnwrapETH` and `ReturnFunds` in the Payback part of `WidtdrawPayback`

### DIFF
--- a/packages/oasis-actions/src/operations/aave/paybackWithdraw.ts
+++ b/packages/oasis-actions/src/operations/aave/paybackWithdraw.ts
@@ -8,6 +8,7 @@ import { AAVEStrategyAddresses } from './addresses'
 export async function paybackWithdraw(args: {
   amountCollateralToWithdrawInBaseUnit: BigNumber
   amountDebtToPaybackInBaseUnit: BigNumber
+  isPaybackAll: boolean
   collateralTokenAddress: string
   collateralIsEth: boolean
   debtTokenAddress: string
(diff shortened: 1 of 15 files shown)


packages/oasis-actions/src/operations/aave/paybackWithdraw.ts
```diff
@@ -34,7 +35,13 @@ export async function paybackWithdraw(args: {
   const paybackDebt = actions.aave.aavePayback({
     asset: args.debtTokenAddress,
     amount: args.amountDebtToPaybackInBaseUnit,
-    paybackAll: false,
+    paybackAll: args.isPaybackAll,
+  })
+  const unwrapEthDebt = actions.common.unwrapEth({
+    amount: new BigNumber(MAX_UINT),
+  })
+  const returnLeftFundFromPayback = actions.common.returnFunds({
```
zerotucks, line 43: could be `returnSurplusFundsAfterPayback`

test/fixtures/factories/createEthUsdcMultiplyAAVEPosition.ts
```diff
@@ -5,22 +5,22 @@ import { executeThroughDPMProxy, executeThroughProxy } from '../../../helpers/de
...
 import { AavePositionStrategy, PositionDetails, StrategiesDependencies } from '../types'
+import { ETH, MULTIPLE, SLIPPAGE, USDC } from './common'
 import { OpenPositionTypes } from './openPositionTypes'
 
-async function getEthUsdcMultiplyAAVEPosition(dependencies: OpenPositionTypes[1]) {
-  const debtToken = { symbol: 'USDC' as const, precision: 6 }
-  const collateralToken = { symbol: 'ETH' as const, precision: 18 }
+const depositCollateralAmount = amountToWei(new BigNumber(10), ETH.precision)
 
+async function getEthUsdcMultiplyAAVEPosition(dependencies: OpenPositionTypes[1]) {
```
zerotucks, line 13: Any reason why we use `get` instead of `open` here?

I feel like there might be a good reason but double checking

zerotucks (approved):

PR #176, opened 2023-01-29
Aave V3 actions

fenris85:
https://app.shortcut.com/oazo-apps/story/7520/2-days-spike-aave-v3-poc-actions

**Changes**
- Add AAVE V3 Actions including setEMode
- Write unit tests to cover new actions


abi/aaveOracle.json
```diff
@@ -0,0 +1 @@
...view","type":"function"},{"inputs":[{"internalType":"address","name":"asset","type":"address"}],"name":"getAssetPrice","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address[]","name":"assets","type":"address[]"}],"name":"getAssetsPrices","outputs":[{"internalType":"uint256[]","name":"","type":"uint256[]"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"getFallbackOracle","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"asset","type":"address"}],"name":"getSourceOfAsset","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address[]","name":"assets","type":"address[]"},{"internalType":"address[]","name":"sources","type":"address[]"}],"name":"setAssetSources","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"address","name":"fallbackOracle","type":"address"}],"name":"setFallbackOracle","outputs":[],"stateMutability":"nonpayable","type":"function"}]
```
peculiarity, line 1: I think, it's better to be renamed to `aavev3PriceOracle` or something like that. Can you check what's the difference between this one and `aavePriceOracle` that we currentl have?

abi/aaveProtocolDataProvider.json
```diff
@@ -0,0 +1 @@
...,"name":"getTotalDebt","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"asset","type":"address"}],"name":"getUnbackedMintCap","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[{"internalType":"address","name":"asset","type":"address"},{"internalType":"address","name":"user","type":"address"}],"name":"getUserReserveData","outputs":[{"internalType":"uint256","name":"currentATokenBalance","type":"uint256"},{"internalType":"uint256","name":"currentStableDebt","type":"uint256"},{"internalType":"uint256","name":"currentVariableDebt","type":"uint256"},{"internalType":"uint256","name":"principalStableDebt","type":"uint256"},{"internalType":"uint256","name":"scaledVariableDebt","type":"uint256"},{"internalType":"uint256","name":"stableBorrowRate","type":"uint256"},{"internalType":"uint256","name":"liquidityRate","type":"uint256"},{"internalType":"uint40","name":"stableRateLastUpdated","type":"uint40"},{"internalType":"bool","name":"usageAsCollateralEnabled","type":"bool"}],"stateMutability":"view","type":"function"}]
```
peculiarity, line 1: Same comment as the above one. We already have `aaveDataProvider`. Might be a good idea to rename it so that we differentiate the two

test/deploySystem.ts
```diff
@@ -175,6 +175,32 @@ export async function deploySystem(config: RuntimeConfig, debug = false, useFall
...
 
+  //-- AAVE V3 Actions
+  // const [depositInAAVEV3Action, actionDepositInAAVEV3Address] = await deploy(
+  //   CONTRACT_NAMES.aaveV3.DEPOSIT,
+  //   [serviceRegistryAddress],
+  // )
+  //
+  // const [borrowInAAVEV3Action, actionAaveV3BorrowAddress] = await deploy(
+  //   CONTRACT_NAMES.aaveV3.BORROW,
+  //   [serviceRegistryAddress],
```
peculiarity, line 186: I think this should be uncommited

abi/aaveOracle.json
zerotucks: Yea, there's another /abi directory in the library.

I'm not loving the two separate abi directories.

Feel like we should have one. Probably in the repo root that is used by tasks and the library? 

Do you have an opinion on that? Once we've decided that I'll update all the abi filenames so they're nice and clear

jakubswierczek: `aaveOracle` is the name from the documentation. It's a kind of official name. Should stay.

PR #185, opened 2023-02-07
Gg/close to collateral

peculiarity:


helpers/serviceRegistry.ts
```diff
@@ -4,6 +4,7 @@ import { Signer, utils } from 'ethers'
 export class ServiceRegistry {
   address: string
   signer: Signer
+  ethers: any
```
jakubswierczek: please add type here

packages/oasis-actions/src/operations/aave/close.ts
```diff
@@ -81,11 +82,19 @@ export async function close(
...
     asset: args.debtTokenIsEth ? ADDRESSES.main.ETH : args.debtTokenAddress,
   })
+  returnDebtFunds.skipped = args.shouldCloseToCollateral
 
   const returnCollateralFunds = actions.common.returnFunds({
     asset: args.collateralIsEth ? ADDRESSES.main.ETH : args.collateralTokenAddress,
   })
 
+  const sendRemainingDebtFundsToFeeRecipient = actions.common.sendToken({
+    asset: args.debtTokenAddress,
```
jakubswierczek, line 92: In that case, we will be sending `WETH.` Maybe we can unwrap it first?

packages/oasis-actions/src/strategies/aave/close.ts
```diff
@@ -110,19 +242,23 @@ async function buildOperation(
     .integerValue(BigNumber.ROUND_DOWN)
 
   const closeArgs = {
-    lockedCollateralAmountInWei: args.collateralAmountLockedInProtocolInWei,
+    lockedCollateralAmountInWei: dependencies.shouldCloseToCollateral
+      ? swapData.fromTokenAmount
+      : args.collateralAmountLockedInProtocolInWei,
     flashloanAmount: amountToFlashloanInWei,
-    fee: DEFAULT_FEE,
+    fee: dependencies.shouldCloseToCollateral ? 0 : DEFAULT_FEE, //TODO - fee should be passed
```
jakubswierczek, line 257: Why is it 0 ?

peculiarity: Because for `closeToCollateral` We ain't going to collect fee. I think it's written in all the comments that I left.

packages/oasis-actions/src/operations/aave/close.ts
peculiarity: That's the thing. We don't want to unwrap it. This is send to the feeBeneficiery address.

jakubswierczek (approved):

jakubswierczek (approved):

packages/oasis-actions/src/strategies/aave/close.ts
```diff
@@ -64,24 +186,36 @@ async function getSwapData(args: AAVECloseArgs, dependencies: AAVECloseDependenc
...
+  colPrice: BigNumber,
+  colPrecision: number,
+  debtAmount: BigNumber,
+  fee: BigNumber,
+  slippage: BigNumber,
 ) {
-  return collectFeeFrom === 'sourceToken'
-    ? calculateFee(swapAmountBeforeFees, new BigNumber(DEFAULT_FEE), new BigNumber(FEE_BASE))
-    : ZERO
+  const paybackAmount = debtPrice.times(debtAmount)
```
zerotucks, line 208:
```suggestion
    // Debt per protocol base currency x DebtAmount = Payback Amount (Units in base currency)
  const paybackAmount = debtPrice.times(debtAmount)
```

packages/oasis-actions/src/strategies/aave/close.ts
```diff
@@ -64,24 +186,36 @@ async function getSwapData(args: AAVECloseArgs, dependencies: AAVECloseDependenc
...
+  const paybackAmountInclFee = paybackAmount.times(ONE.plus(fee))
+  const collateralAmountNeeded = new BigNumber(
```
zerotucks, line 212:
```suggestion
  // Payback Amount (Units in protocol base currency) / Collateral Price (In protocol base currency) = Collateral Amount Needed (Units in protocol base currency)
  const collateralAmountNeeded = new BigNumber(
```

packages/oasis-actions/src/strategies/aave/close.ts
zerotucks: If you think worth it. Just wrote comment in response to existing comment that is should be double checked

packages/oasis-actions/src/strategies/aave/close.ts
zerotucks: Do we need the TODO?

zerotucks (approved):

PR #192, opened 2023-02-10
fix: fees issue causing dust balances to remain on proxy

zerotucks:
Incorrect balance being approved on Swap.sol relative to amount being swapped by 1inch.

### DIFF
--- a/packages/oasis-actions/src/strategies/aave/open/open.ts
+++ b/packages/oasis-actions/src/strategies/aave/open/open.ts
@@ -23,6 +23,8 @@ import {
   AaveVersion,
 } from '../getCurrentPosition'
 
+type WithFee = { fee: BigNumber }
+
 export interface AaveOpenArgs {
   depositedByUser?: {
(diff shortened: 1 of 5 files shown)


packages/oasis-actions/src/strategies/aave/open/open.ts
```diff
@@ -137,7 +148,7 @@ export async function open(
 
 async function simulatePositionTransition(
   quoteSwapData: SwapData,
-  args: AaveOpenArgs,
+  args: AaveOpenArgs & { fee: BigNumber },
```
jakubswierczek: Maybe we can create a type:
```
type WithFee = { fee: BigNumber }
```

Or even:
```
type WithFee = { amount: BigNumber; token: string }
```
Then we can return `IPositionTranstion & WithFee` because the fee is now bounded with the Swap. It could be a better approach because, on Close To Collateral, we take fee after payback debt.

jakubswierczek (approved):

PR #208, opened 2023-02-23
AAVE V3 Adjust Strategy & Operations

zerotucks:
**How to test**
From root directory run `yarn hardhat test test/aave/Adjust.test.ts`

**Changes**
- Creates V3 AdjustRiskUp and AdjustRiskDown operations
- Creates shared strategy logic for Adjusting Up and Down for both versions
- Exposes this strategy from the library via aave.v2.adjust and aave.v3.adjust
- Does away with logic to further refine the Final Position returned by the strategy to use a marginally more accurate swap price. Was causing some complications so removed because it's more complexity for marginal gain
- Having created Types for the Operation args that can be composed together. The idea being that we use similar structured arg types and compose them together to get what we need. This was a test to see how it worked as would like to apply the same thinking to the strategies themselves with the goal being to establish some uniformity between strategy args.


packages/oasis-actions/src/strategies/aave/adjust/adjust.ts
```diff
@@ -0,0 +1,729 @@
...
+  })
+
+  // SimulateAdjustUp
+  const { simulatedPositionTransition: simulatedAdjustUp, reserveEModeCategory } =
+    await simulatePositionTransition(
+      isAdjustUp,
+      quoteSwapData,
+      { ...args, fee },
+      dependencies,
+      true,
```
jakubswierczek, line 102: remove debug

jakubswierczek (changes requested):
Disable unnecessary debugs.

packages/oasis-actions/src/strategies/aave/adjust/adjust.ts
zerotucks: That's a flag for something else.

We could replace with named args to make more clear or just merge.

jakubswierczek (approved):

PR #211, opened 2023-02-28
Add Chainlog viewer contract and script to test it

fenris85:
Add `ChainLogView` contract to read Maker services addresses from it instead of our own `ServiceRegistry`


contracts/actions/common/TakeFlashloan.sol
```diff
@@ -7,20 +7,23 @@ import { IERC3156FlashBorrower } from "../../interfaces/flashloan/IERC3156FlashB
...
  * @title TakeFlashloan Action contract
  * @notice Executes a sequence of Actions after flashloaning funds
  */
 contract TakeFlashloan is Executable, ProxyPermission {
   ServiceRegistry internal immutable registry;
+  ChainLogView internal immutable chainlogView;
   address internal immutable dai;
 
-  constructor(ServiceRegistry _registry, address _dai) {
+  constructor(ServiceRegistry _registry, address _dai, ChainLogView _chainlogView) {
```
peculiarity, line 23: If `registry` is not used anywhere then just remove it. No need to provide additional argument during deployment and spend gas on the storage variable for it .

peculiarity (approved):

PR #221, opened 2023-03-07
Create L2 actions for AAVEv3. Integrate Balance Flashloan provider

peculiarity:
- Optimism Fork Configuration - Additional params in the `.env` file
  -  `NETWORK_FORK`- `Optimism` or `Mainnet`
  -  `OPTIMISM_BLOCK_NUMBER` - Block number to fork from on Optimism
  -  `OPTIMISM_URL` - Url to a node on optimism
- AAVEv3 Actions for Optimism
- TakeFlashloan Action now can operation over 2 Flashloan provider ( Maker's Flash Mint Module and Balancer )
- OperationExecutor now handles flashloans from both Maker's Flash Mint Module and Balancer
- 
**Note:*** Don't check the `DepositV3.test.ts, PaybackV3.test.ts, WithdrawV3.test.ts, BorrowV3.test.ts, SetModeV3.test.ts` - Just moved to a new folder and that's it.

**Note:** Most likely the library will break + all the test around it. The data struct  that is encoded for the `TakeFlashloan` action is changed and these changes should be reflected in the library as well. 

New data struct looks like this: 

```
struct FlashloanData {
  uint256 amount;
  address asset;
  bool isProxyFlashloan;
  bool isDPMProxy;
  FlashloanProvider provider;
  Call[] calls;
}
```

- `asset` - The address of the asset that will be flashloaned. It will be ignored if the flashloan provider is selected to be Maker's Flash Mint Module

- `provider` - That is a value between either 0 or 1 for now. Depending on which prodiver we are choosing.


contracts/core/OperationExecutor.sol
```diff
@@ -120,16 +126,74 @@ contract OperationExecutor is IERC3156FlashBorrower {
     bytes calldata data
   ) external override returns (bytes32) {
     address lender = registry.getRegisteredService(FLASH_MINT_MODULE);
+    FlashloanData memory flData = abi.decode(data, (FlashloanData));
+
+    checkIfLenderIsTrusted(lender);
+    checkIfFlashloanedAssetIsTheRequiredOne(asset, flData.asset);
+    checkIfFlashloanedAmountIsTheRequiredOne(asset, flData.amount);
```
zerotucks, line 133:
```suggestion
    checkIfFlashloanedAmountIsTheRequiredOne(amount, flData.amount);
```

fenris85 (approved):

contracts/core/OperationExecutor.sol
```diff
@@ -120,16 +126,74 @@ contract OperationExecutor is IERC3156FlashBorrower {
...
 
-    require(IERC20(asset).balanceOf(address(this)) >= flData.amount, "Flashloan inconsistency");
+    checkIfLenderIsTrusted(lender);
+    checkIfFlashloanedAssetIsTheRequiredOne(asset, flData.asset);
+    checkIfFlashloanedAmountIsTheRequiredOne(asset, flData.amount);
 
-    if (flData.isProxyFlashloan) {
-      IERC20(asset).safeTransfer(initiator, flData.amount);
+    // Can we just assume that the initiator is the OperationExecutor?
+    // Are there any security consequences because of that assumption?
```
zerotucks, line 162: Not sure if we have an alternative with Balancer anyway.

hardhat.config.ts
```diff
@@ -37,17 +37,55 @@ task('accounts', 'Prints the list of accounts', async (taskArgs, hre) => {
     console.log(account.address)
   }
 })
+const networkFork = process.env.NETWORK_FORK
 
-const blockNumber = process.env.BLOCK_NUMBER
-if (!blockNumber) {
-  throw new Error(`You must provide a block number.`)
+if (!networkFork && (networkFork == 'Mainnet' || networkFork == 'Optimism')) {
```
zerotucks:
```suggestion
if (!networkFork || !(networkFork == 'Mainnet' || networkFork == 'Optimism')) {
```
Was that the intention?

zerotucks (approved):

contracts/core/OperationExecutor.sol
peculiarity: `asset` is the correct  one. WE pass the asset to check the contract balanace and see if's more than the required one ( `flData.amount `

contracts/core/OperationExecutor.sol
peculiarity: WEll that's the thing. I just assumed that it will be always the Executor,. IF you can't think of any security issues right there then it's fine I guess :D

contracts/core/OperationExecutor.sol
zerotucks: Right got it!!

PR #224, opened 2023-03-10
Addresses cleanup [WIP]

fenris85:


zerotucks:
Think we can probably close this one

PR #227, opened 2023-03-13
Aave V3 L2 Open Operations

zerotucks:
**Description**
Get L2 Open operations working with Optimism

**Notable Changes & Temporary items**
* ProxyPermission - is using a hardcoded L2 factory address. This probably needs to come from Registry. Implication being we can't easily swap back and forth between L1 and L2 tests
* L2 Tests use v5.0  of 1inch api and are configured with v5.0 1inch router in test config
* Inside Flashloan Action and Operation Executor we're augmenting the flashloan data we get from the action to include the initiator's address. Opted not to pass the initiator address (proxy) from the library.


hardhat.config.ts
```diff
@@ -114,15 +116,15 @@ const config: HardhatUserConfig = {
     local: {
       url: 'http://127.0.0.1:8545',
       timeout: 1000000,
-      chainId: 2137,
+      chainId: ChainIdByNetwork[Network.LOCAL],
     },
     hardhat: {
       forking: {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         url: forkConfig ? forkConfig.nodeURL : 'http:127.0.01:8545',
```
peculiarity: I don't know why it is like that but the IP here is messed up. Can you also fix it in here please?

contracts/core/OperationExecutor.sol
```diff
@@ -191,7 +192,7 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
     if (assetBalance < requiredAmount) revert InconsistentAmount(assetBalance, requiredAmount);
   }
 
-  function processFlashloan(FlashloanData memory flData, address initiator) private {
+  function processFlashloan(FlashloanWithInitiatorData memory flData, address initiator) private {
```
peculiarity: If `FlashloanWithInitiatorData` is passed, then there is no point of having a separate param called `initiator`. This value is already contained in the `flData`.

packages/oasis-actions/src/operations/aave/v2/adjustRiskUp.ts
```diff
@@ -128,10 +128,11 @@ export async function adjustRiskUp({
   ]
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
+    isDPMProxy: proxy.isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: flashloan.amount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
-    isDPMProxy: proxy.isDPMProxy,
+    provider: new BigNumber(0),
```
peculiarity: Would be nice to have an enum or some constant with the providers. Something like
```
cont FlashloanProviders = {
  DssFlash: ZERO,
  Balancer: ONE,
}
```
Right now these providers are like an enum in the contract sand correspond to numbers but at ay point of time it could be a `provider` configuration or could be replaced with a string.

Abstracting it like that you will change the value  in the definition only and  you should worry about where it's been used.

**Edit**: Actually I wrote that comment before I saw you actually defined such a constant.
1st. Why isn't it used everywhere?
2nd. What do you have to wrap it into a BigNumber?

If it needs to be a BigNumber, just use `ZERO` and `ONE` in the definition instead of wrapping it everywhere.

packages/oasis-actions/src/operations/aave/v2/open.ts
```diff
@@ -201,9 +201,10 @@ export async function open({
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
     isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: flashloanAmount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
+    provider: new BigNumber(0),
```
peculiarity: Same as the above comment for the value

packages/oasis-actions/src/operations/aave/v3/adjustRiskDown.ts
```diff
@@ -111,9 +111,10 @@ export async function adjustRiskDown({
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
     isDPMProxy: proxy.isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: flashloan.amount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
+    provider: new BigNumber(flashloan.provider),
```
peculiarity: Unify them with the previous comments

packages/oasis-actions/src/operations/aave/v3/close.ts
```diff
@@ -148,10 +148,11 @@ export async function close(
   unwrapEth.skipped = !args.debtTokenIsEth && !args.collateralIsEth
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
+    isDPMProxy: args.isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: args.flashloanAmount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
-    isDPMProxy: args.isDPMProxy,
+    provider: new BigNumber(0),
```
peculiarity: Same

packages/oasis-actions/src/operations/aave/v3/adjustRiskUp.ts
```diff
@@ -117,9 +117,10 @@ export async function adjustRiskUp({
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
     isDPMProxy: proxy.isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: flashloan.amount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
+    provider: new BigNumber(flashloan.provider),
```
peculiarity: Same

packages/oasis-actions/src/operations/aave/v3/addresses.ts
```diff
@@ -9,5 +9,5 @@ export interface AAVEV3StrategyAddresses {
   chainlinkEthUsdPriceFeed: string
   aaveOracle: string
   pool: string
-  aaveProtocolDataProvider: string
+  poolDataProvider: string
```
peculiarity, line 12: Any particular reason for changing the name?

packages/oasis-actions/src/operations/aave/v2/adjustRiskDown.ts
```diff
@@ -92,10 +92,11 @@ export async function adjustRiskDown({
   ]
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
+    isDPMProxy: proxy.isDPMProxy,
+    asset: addresses.DAI,
     flashloanAmount: flashloan.amount,
-    borrower: addresses.operationExecutor,
     isProxyFlashloan: true,
-    isDPMProxy: proxy.isDPMProxy,
+    provider: new BigNumber(0),
```
peculiarity: Consider the suggested change

packages/oasis-actions/src/helpers/flashloan/resolve-provider.ts
```diff
@@ -0,0 +1,30 @@
+import { providers } from 'ethers'
+
+import { Network, NetworkByChainId } from '../../../../../helpers/network'
+import { FlashloanProvider } from '../../types/common'
+
+export const resolveFlashloanProvider = async (provider: providers.Provider) => {
```
peculiarity: I don't think this should be here at all. This code more or less is in the `network.ts`

packages/oasis-actions/src/helpers/flashloan/resolve-provider.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!(localProvider as providers.JsonRpcProvider).send)
+      throw new Error('Provider does not support send method')
+    const metadata = await localProvider.send('hardhat_metadata', [])
+    return resolveProvider(NetworkByChainId[metadata.forkedNetwork.chainId])
+  }
+
+  return resolveProvider(NetworkByChainId[chainId])
+}
+
+function resolveProvider(network: Network): FlashloanProvider {
```
peculiarity: All you need this is this function renamed to `resolveFlashloanProvider`

packages/oasis-actions/src/operations/aave/v3/open.ts
```diff
@@ -85,78 +81,82 @@ export const operationDefinition = {
...
-  isDPMProxy,
-  positionType,
+  position,
+  emode,
+  addresses,
 }: OpenArgs): Promise<IOperation> {
+  const depositAmount = deposit?.amount || ZERO
+  const depositAddress = deposit?.address || NULL_ADDRESS
+
+  const isDepositingCollateral = deposit?.address === collateral.address
```
peculiarity, line 97: You can use `depositAmount` here which has been already checked on `line:94`.

packages/oasis-actions/src/operations/aave/v3/open.ts
```diff
@@ -85,78 +81,82 @@ export const operationDefinition = {
...
+  const isDepositingDebtTokens = deposit?.address === debt.address
```
peculiarity, line 98: Same as the above

packages/oasis-actions/src/operations/aave/v3/open.ts
```diff
@@ -206,10 +206,11 @@ export async function open({
...
 
   const takeAFlashLoan = actions.common.takeAFlashLoan({
-    isDPMProxy,
-    flashloanAmount: flashloanAmount,
-    borrower: addresses.operationExecutor,
+    isDPMProxy: proxy.isDPMProxy,
+    asset: addresses.DAI,
+    flashloanAmount: flashloan.amount,
     isProxyFlashloan: true,
+    provider: new BigNumber(flashloan.provider),
```
peculiarity: Been commented on this

packages/oasis-actions/src/protocols/aave/getAaveProtocolData.ts
```diff
@@ -172,3 +187,26 @@ async function getAaveV3ProtocolData({
 }
 
 export type AaveProtocolData = ReturnType<typeof getAaveProtocolData>
+
+type AllowedContractNames = 'poolDataProvider' | 'pool' | 'aaveOracle'
+async function getAbiForContract(contractName: AllowedContractNames, provider: providers.Provider) {
+  const network = await getForkedNetwork(provider as providers.JsonRpcProvider)
```
peculiarity: This is mesleading. If we operation on mainnet or optimism or any network not through hardhat... it's not a forked network.

peculiarity (changes requested):

contracts/actions/common/TakeFlashloan.sol
```diff
@@ -40,7 +44,17 @@ contract TakeFlashloan is Executable, ProxyPermission {
...
+    FlashloanData memory flDataFromProps = parseInputs(data);
+    // We need to pass the initiator address to the Operation Executor so that it can re-establish the calling context
+    FlashloanWithInitiatorData memory flData = FlashloanWithInitiatorData({
+      amount: flDataFromProps.amount,
+      asset: flDataFromProps.asset,
+      isProxyFlashloan: flDataFromProps.isProxyFlashloan,
+      isDPMProxy: flDataFromProps.isDPMProxy,
+      provider: flDataFromProps.provider,
+      calls: flDataFromProps.calls,
+      initiator: address(this)
```
peculiarity, line 56: Instead of copying the data, we can pass this struct with a `zero` address as initiator from the UI and when it gets here it's gong to be updated. Any thoughts

contracts/actions/common/TakeFlashloan.sol
```diff
@@ -40,7 +44,17 @@ contract TakeFlashloan is Executable, ProxyPermission {
    * @param data Encoded calldata that conforms to the FlashloanData struct
    */
   function execute(bytes calldata data, uint8[] memory) external payable override {
-    FlashloanData memory flData = parseInputs(data);
+    FlashloanData memory flDataFromProps = parseInputs(data);
+    // We need to pass the initiator address to the Operation Executor so that it can re-establish the calling context
+    FlashloanWithInitiatorData memory flData = FlashloanWithInitiatorData({
```
peculiarity: We don't need this whole new structure. You can encode the initiatero with the `flData` directly.

contracts/actions/common/TakeFlashloan.sol
```diff
@@ -55,7 +69,7 @@ contract TakeFlashloan is Executable, ProxyPermission {
         IERC3156FlashBorrower(operationExecutorAddress),
         dai,
         flData.amount,
-        data
+        abi.encode(flData)
```
peculiarity: `abi.encode(flData, address(this))`

contracts/core/OperationExecutor.sol
```diff
@@ -126,7 +127,7 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
     bytes calldata data
   ) external override returns (bytes32) {
     address lender = registry.getRegisteredService(FLASH_MINT_MODULE);
-    FlashloanData memory flData = abi.decode(data, (FlashloanData));
+    FlashloanWithInitiatorData memory flData = abi.decode(data, (FlashloanWithInitiatorData));
```
peculiarity: ```(FlashloanData memory flData, address initiator) = abi.decode(data, (FlasloanData, address))```

contracts/core/OperationExecutor.sol
```diff
@@ -152,15 +153,15 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
...
 
     checkIfLenderIsTrusted(lender);
     checkIfFlashloanedAssetIsTheRequiredOne(asset, flData.asset);
     checkIfFlashloanedAmountIsTheRequiredOne(asset, flData.amount);
 
     // Can we just assume that the initiator is the OperationExecutor?
     // Are there any security consequences because of that assumption?
-    processFlashloan(flData, registry.getRegisteredService(OPERATION_EXECUTOR));
```
peculiarity, line 164: That was completely wrong indeed. It worked on my case because I was executing the flow directly and not through a proxy.  Indeed the proxy is the initiator. Can you check the test for L2 though ? They might be broken and if some, let me know and I can work on them

packages/oasis-actions/src/helpers/flashloan/resolve-provider.ts
zerotucks: We do need this because when running a local node you might be forking either mainnet or optimism.

So, returning just DSSFlash for local node would be insufficient.

packages/oasis-actions/src/operations/aave/v3/open.ts
zerotucks: I was trying to preserve the existing two Pull Actions so we don't have to redeploy the Open operations on L1

If I get rid of that check and just have 1 Pull Action we'll need to deploy. So have left for now.

peculiarity: I don't follow this to be honest.  On line:95 you have `const depositAmount = deposit?.address || NULL_ADDRESS`.
Why can't you use `depositAddress` on line ` const isDepositingCollateral = depositAddress === collateral.address`?

contracts/core/OperationExecutor.sol
```diff
@@ -17,7 +17,7 @@ import { FlashloanData, Call } from "./types/Common.sol";
 import { OPERATION_STORAGE, OPERATIONS_REGISTRY, OPERATION_EXECUTOR } from "./constants/Common.sol";
 import { FLASH_MINT_MODULE } from "./constants/Maker.sol";
 import { BALANCER_VAULT } from "./constants/Balancer.sol";
-
+import "hardhat/console.sol";
```
peculiarity: Remove that please

contracts/core/OperationExecutor.sol
```diff
@@ -106,14 +107,13 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
   }
 
   /**
-   * @notice Not to be called directly.
+       * @notice Not to be called directly.
```
peculiarity: No need of that I guess

contracts/actions/common/TakeFlashloan.sol
```diff
@@ -23,7 +23,11 @@ contract TakeFlashloan is Executable, ProxyPermission {
   ServiceRegistry internal immutable registry;
   address internal immutable dai;
 
-  constructor(ServiceRegistry _registry, address _dai) {
+  constructor(
+    ServiceRegistry _registry,
+    address _dai,
+    address _dsGuardFactory
```
peculiarity, line 29: There is this param for the construct `_dsGuardFactory` which is not used at all

test/fixtures/factories/createEthUsdcMultiplyAAVEPosition.ts
```diff
@@ -12,19 +12,20 @@ import {
   aaveV2UniqueContractName,
   aaveV3UniqueContractName,
 } from '../../../packages/oasis-actions/src/protocols/aave/config'
+import { DeploymentSystem } from '../../../scripts/deployment20/deploy'
 import { AavePositionStrategy, PositionDetails, StrategiesDependencies } from '../types'
 import { ETH, MULTIPLE, SLIPPAGE, UNISWAP_TEST_SLIPPAGE, USDC } from './common'
 import { OpenPositionTypes } from './openPositionTypes'
 
-const depositCollateralAmount = amountToWei(new BigNumber(10), ETH.precision)
+const depositCollateralAmount = amountToWei(new BigNumber(1), ETH.precision)
```
peculiarity: there is a constant `ONE`

test/fixtures/factories/common.ts
```diff
@@ -1,23 +1,33 @@
...
 export const UNISWAP_TEST_SLIPPAGE = new BigNumber(0.25)
 export const SLIPPAGE = new BigNumber(0.05)
 export const MULTIPLE = new BigNumber(2)
 export const EMODE_MULTIPLE = new BigNumber(9.5)
 
-export const USDC: TokenDetails = {
-  symbol: 'USDC' as const,
-  precision: 6,
-  address: mainnetAddresses.USDC,
+export class USDC {
```
peculiarity, line 11: Do we need to wrap the tokens into classes :S ?

test/addresses/optimism.ts
```diff
@@ -0,0 +1,19 @@
+import { ADDRESSES } from '@oasisdex/oasis-actions/src/helpers/addresses'
+
+export const optimismAddresses = {
```
peculiarity, line 3: Don't we have a single place with all the needed addresses? Why add a new place again ?

scripts/deployment20/test-configs/test-aave-v3-mainnet.conf.json
```diff
@@ -96,7 +96,8 @@
         "history": [],
         "constructorArgs": [
           "address:ServiceRegistry",
-          "0x6B175474E89094C44Da98b954EedeAC495271d0F"
+          "0x6B175474E89094C44Da98b954EedeAC495271d0F",
+          "0x5a15566417e6C1c9546523066500bDDBc53F88C7"
```
peculiarity, line 87: This won't be needed because the constructor is wrong ( there is a param that's not used at all )

packages/oasis-actions/src/helpers/flashloan/resolve-provider.ts
peculiarity: My point is that you already have more or less the same code in `network.ts`. There should be one place responsible for the network resolution.

contracts/actions/common/TakeFlashloan.sol
peculiarity: So cna we remove this at the end ? I'm confused

contracts/actions/common/TakeFlashloan.sol
```diff
@@ -28,7 +32,7 @@ contract TakeFlashloan is Executable, ProxyPermission {
   ServiceRegistry internal immutable registry;
   address internal immutable dai;
 
-  constructor(ServiceRegistry _registry, address _dai) {
+  constructor(ServiceRegistry _registry, address _dai, address _dsGuardFactory) ProxyPermission(_dsGuardFactory) {
```
peculiarity: You don't need this

peculiarity: Why not pass the `ServiceRegistry` as a dependency isntead of `dsGuardFactory` address?

test/addresses/optimism.ts
zerotucks: In dev there's already test/addresses/mainnet.ts so this is just following on from that pattern is essentially flattening some of the addresses that come the ADDRESSES mapping

test/fixtures/factories/common.ts
zerotucks: We have these options...:
1. Turn them into classes so that the address can be configured.
2. Create separate token helpers for different networks
3. Remove address from the token helpers that already exist

hardhat.config.ts
```diff
@@ -118,14 +120,15 @@ const config: HardhatUserConfig = {
...
       url: 'http://127.0.0.1:8545',
       timeout: 1000000,
-      chainId: 2137,
+      chainId: ChainIdByNetwork[Network.LOCAL],
     },
     hardhat: {
       forking: {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        url: forkConfig ? forkConfig.nodeURL : 'http:127.0.01:8545',
+        url: forkConfig ? forkConfig.nodeURL : 'http://127.0.01:8545',
```
peculiarity: "." between `0` and `1`

peculiarity (approved):

PR #228, opened 2023-03-14
Ajna earn validations

piekczyk:
Refactored handling of ajna earn validations and methods

### DIFF
--- a/packages/oasis-actions/src/helpers/ajna/index.ts
+++ b/packages/oasis-actions/src/helpers/ajna/index.ts
@@ -1,0 +1,106 @@
+import BigNumber from 'bignumber.js'
+import { ethers } from 'ethers'
+
+import poolAbi from '../../../../../abi/external/ajna/ajnaPoolERC20.json'
+import { getAjnaValidations } from '../../strategies/ajna/earn/validations'
+import { AjnaEarnPosition } from '../../types/ajna'
+import { AjnaPool } from '../../types/ajna/AjnaPool'
+import {
+  Address,
+  AjnaDependencies,
+  AjnaEarnActions,
+  AjnaError,
+  AjnaWarning,
+  Strategy,
+} from '../../types/common'
+
+export interface AjnaEarnArgs {
+  poolAddress: Address
+  dpmProxyAddress: Address
+  quoteAmount: BigNumber
+  quoteTokenPrecision: number
+  price: BigNumber
+  position: AjnaEarnPosition
+}
+
+export const prepareAjnaPayload = <T extends { pool: AjnaPool }>({
+  dependencies,
+  args,
+  targetPosition,
+  errors,
+  warnings,
+  data,
+}: {
+  dependencies: AjnaDependencies
+  targetPosition: T
+  errors: AjnaError[]
+  warnings: AjnaWarning[]
+  data: string
+  args: {
+    position: T
+    quoteAmount: BigNumber
+    quoteTokenPrecision: number
+  }
+}): Strategy<T> => {
+  const isDepositingEth =
+    args.position.pool.collateralToken.toLowerCase() === dependencies.WETH.toLowerCase()
+
+  return {
+    simulation: {
+      swaps: [],
+      errors,
+      warnings,
+      targetPosition,
+      position: targetPosition,
+    },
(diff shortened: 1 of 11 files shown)


packages/oasis-actions/src/helpers/ajna/index.ts
```diff
@@ -0,0 +1,106 @@
...
+}) => {
+  const pool = new ethers.Contract(args.poolAddress, poolAbi, dependencies.provider)
+  const [poolDebt] = await pool.debtInfo()
+
+  const afterLupIndex =
+    action === 'withdraw'
+      ? await pool.depositIndex(
+          new BigNumber(poolDebt.toString()).plus(args.quoteAmount.shiftedBy(18)).toString(),
+        )
+      : undefined
```
paszkowskiDamian, line 87: In case of deposit the LUP can also change, for sure it will not prevent us from executing an action but maybe it can be interesting to see it.

piotrkonowrocki (approved):

packages/oasis-actions/src/helpers/ajna/index.ts
paszkowskiDamian: Maybe after moving contracts to AjnaPosition we could do the LUP simulation on the deposit method 🤔

packages/oasis-actions/src/helpers/ajna/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      errors,
+      warnings,
+      targetPosition,
+      position: targetPosition,
+    },
+    tx: {
+      to: dependencies.ajnaProxyActions,
+      data,
+      value: isDepositingEth
+        ? ethers.utils.parseUnits(args.quoteAmount.toString(), args.quoteTokenPrecision).toString()
```
paszkowskiDamian: What do you think about using `parsEther` instead of `parseUnits`? In here quoteToken can only be ETH.

packages/oasis-actions/src/strategies/ajna/earn/depositAndAdjust.ts
```diff
@@ -53,7 +38,8 @@ export async function depositAndAdjust(
     .then((res: any) => res.toString())
     .then((res: string) => new BigNumber(res))
 
-  let data: string | null = null
+  let data = ''
+  let targetPosition: AjnaEarnPosition | null = null
```
paszkowskiDamian, line 42: would it be easer for you if you would assign current position to target position?

packages/oasis-actions/src/strategies/ajna/earn/open.ts
```diff
@@ -61,42 +60,34 @@ export async function open(
...
     .then((res: string) => new BigNumber(res))
 
-  const data = await ajnaProxyActions.interface.encodeFunctionData(
+  const data = ajnaProxyActions.interface.encodeFunctionData(
     args.isStakingNft ? 'openEarnPositionNft' : 'openEarnPosition',
     [
       args.poolAddress,
       ethers.utils.parseUnits(args.quoteAmount.toString(), args.quoteTokenPrecision).toString(),
-      ethers.utils.parseUnits(args.price.toString(), 3).toString(),
+      args.price.shiftedBy(18).toString(),
```
paszkowskiDamian, line 72: are you sure it's accepting price with precision 18?

packages/oasis-actions/src/strategies/ajna/earn/validations.ts
```diff
@@ -0,0 +1,67 @@
...
+      name: 'price-below-htp',
+    })
+  }
+
+  // action specific
+  switch (action) {
+    case 'open':
+    case 'deposit': {
+      break
+    }
```
paszkowskiDamian, line 41: ~We can add warning that deposit us under HTP, so it will not earn fees~

sorry we have that already 😅

packages/oasis-actions/src/strategies/ajna/earn/open.ts
piekczyk: yes, we needed this precision and we adjusted AjnaProxyActions in order to handle it

paszkowskiDamian (approved):

PR #233, opened 2023-03-17
Change config files to .ts type

fenris85:


scripts/deployment20/deploy.ts
```diff
@@ -318,33 +310,27 @@ export class DeploymentSystem extends DeployedSystemHelpers {
   }
 
   async deployCore() {
+
+    console.log('DEPLOYING CORE', this.config)
+    console.log('DEPLOYING CORE', this.config.mpa)
```
peculiarity: I guess that was used for debuggging ?

peculiarity (approved):
I think you should remove the `console.logs`s and you are good to go

PR #245, opened 2023-03-24
Integrate new ts configs with existing tests

zerotucks:
**Fixes**
* Update Fixtures to work across networks where necessary
* Add test-configs for both mainnet and optimism that deploy the full system locally
* Populate existing addresses for mainnet as reference in the main mainnet config
* Restrict 1inch routes on Optimism such that Balancer is not a liquidity provider - to avoid triggering reentrancy errors when using Balancer as an FL provider
* Type up the new DeploySystemHelper so it's easier to see the DeploySystem shape returned by the helper, and therefore easier to integrate with existing tests


helpers/swap/OneInchCall.ts
```diff
@@ -54,3 +54,23 @@ export const getOneInchCall =
       exchangeCalldata: response.tx.data,
     }
   }
+
+export const optimismLiquidityProviders = [
```
jakubswierczek, line 58: Do we need that? As we don't provider providers in Mainnet, we shouldn't provide in optimism.

scripts/common/config-item.ts
```diff
@@ -0,0 +1,23 @@
+import { ContractNames } from '@oasisdex/oasis-actions/src'
+import { Address } from '@oasisdex/oasis-actions/src/types'
+
+export type ConfigItem = {
+  name: string
+  deploy?: boolean
```
jakubswierczek: could we remove the optional sign?

scripts/common/config-item.ts
```diff
@@ -0,0 +1,23 @@
...
+  serviceRegistryName?: ContractNames | ''
```
jakubswierczek: I prefer either an optional value or an empty string, but not both.

scripts/common/config-item.ts
```diff
@@ -0,0 +1,23 @@
...
+  address: Address
+  history?: Address[]
```
jakubswierczek: maybe empty array rather than optional?

jakubswierczek (commented):
It looks nice. The `deploy` seems difficult to understand easily. I understand that the deploy generates new `conf` files. Do we want to run it on the merge of the dev branch? In a pipeline? It would generate a new commit, I think.

helpers/swap/OneInchCall.ts
zerotucks: This is so we can exclude Balancer. Because we use Balancer as Flashloan Provider we cannot use Balancer as a 1inch route without throwing a re-entrancy error.

scripts/common/config-item.ts
zerotucks: Cleaned this up a bit. Two config types now: SystemConfigItem and ConfigItem

scripts/common/config-item.ts
zerotucks: As above

scripts/common/config-item.ts
zerotucks: Yes. Good idea. Created a new type

jakubswierczek (approved):

PR #248, opened 2023-03-30
New deployment configuration with the newest optimism addresses

peculiarity:


zerotucks (commented):
Any reason why changing from opt-mainnet to optimism for the conf?

peculiarity:
> Any reason why changing from opt-mainnet to optimism for the conf?

Yes, there is a specific reason. I have no idea why it was named `opt-mainnet`.  The network defined in the `hardhat.config.ts` is called `optimism`. So when we are trying to find a config, we ar elooking for a file called `${network}.conf`. So if our network is called `optimism` we can't have `opt-mainnet`. Also in `hardhat-etherscan` it's called `optimisticEthereum` which is insane and the goerli is `optimisticGoerli` I think. For me `optimism` is fine nad `optimism-goerli` or `optimism-sepholia` or whatever is better.

zerotucks (approved):

PR #249, opened 2023-03-31
Aave v3 L2 close operation

zerotucks:
**Changes**
- Updated Close/Open/CloseToColl tests to work across networks (Mainnet and Optimism)
- Updated test fixtures to work across networks where required
- Update utilities to work across Networks
- Cleaned up unused utilities because when else do they get cleaned
- Aligned Close Operation with Open Operation stylistically
- Updated the CloseToCollateral logic to work on Optimism
- Typed up the DeploySystem and DeployedSystem
- Removed setupLocalSystem from deploy.ts so that it can be removed from all tests and replaced with GG approach
- Typed up deploy config files
- Created a test addresses utility to return flattened test addresses - removed addresses directory in /tests to try and begin to simplify how addresses are handled/represented in repo

**How to test**
* `yarn hardhat test test/e2e/strategies/aave/Close.test.ts`
* `yarn hardhat test test/e2e/strategies/aave/CloseToCollateral.test.ts`

Switch out the NETWORK_FORK (mainnet <> opt-mainnet) environment variable in .env to switch networks


helpers/aave/buildGetTokenByImpersonateFunction.ts
```diff
@@ -1,31 +1,34 @@
...
 import BigNumber from 'bignumber.js'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 
 import erc20abi from '../../abi/external/IERC20.json'
-import { mainnetAddresses } from '../../test/addresses/mainnet'
+import { addressesByNetwork } from '../../test/test-utils/addresses'
 import { RuntimeConfig } from '../types/common'
 
 export type AAVETokensToGet = Exclude<AAVETokens, 'ETH' | 'WETH'>
+const mainnetAddressesForTests = addressesByNetwork(Network.MAINNET)
```
peculiarity, line 12: Why is this only for `mainnet` ?

helpers/aave/buildGetTokenFunction.ts
```diff
@@ -1,29 +1,48 @@
...
     const BUFFER_FACTOR = 1.1
     const amountInInWeth = amount.times(BUFFER_FACTOR).toFixed(0)
+
+    let addresses: NetworkAddressesForTests | undefined
+    if (isMainnetByNetwork(network)) {
+      addresses = addressesByNetwork(Network.MAINNET)
+    }
+    if (isOptimismByNetwork(network)) {
+      addresses = addressesByNetwork(Network.OPT_MAINNET)
+    }
```
peculiarity:
```suggestion
    let addresses:  = addressesByNetwork(network)
```

So you are marking that the return type is of type `NetworkAddressesForTests`. Calling with a specific network type doesn't return any different value than calling the method with just `network`.

test/test-utils/addresses.ts
```diff
@@ -0,0 +1,68 @@
...
+// Type guards
+export function isMainnetByNetwork(network: Network): network is Network.MAINNET {
+  return network === Network.MAINNET
+}
+export function isOptimismByNetwork(network: Network): network is Network.OPT_MAINNET {
+  return network === Network.OPT_MAINNET
+}
+
+export function addressesByNetwork(network: Network.MAINNET): MainnetAddresses
+export function addressesByNetwork(network: Network.OPT_MAINNET): OptMainnetAddresses
```
peculiarity:
```suggestion

```

Don't see much point of this in the way how you use the `addressesByNetwork`

helpers/types/common.ts
```diff
@@ -2,7 +2,9 @@ import { HardhatEthersHelpers } from '@nomiclabs/hardhat-ethers/types'
 import { ethers, providers } from 'ethers'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 
-export type ValueOf<T> = T[keyof T]
+export type PartialRecord<K extends keyof any, T> = {
+  [P in K]?: T
+}
```
peculiarity:
```suggestion

```
There is absolutely no need of that. You can achieve the same thing with the native TS types:
`Partial<Record<K,V>>`

scripts/common/config-item.ts
```diff
@@ -1,26 +1,98 @@
...
+  | CommonContractNames
+  | AaveProtocolContractNames
+
+export type LocalSystemContractNames = 'DsProxyRegistry'
+export type DeployedSystemContractNames =
+  | CoreContractNames
+  | ActionContractNames
+  | LocalSystemContractNames
+
+type SwapRecord = PartialRecord<SwapContractNames, SystemConfigItem>
```
peculiarity:
```suggestion
type SwapRecord = Partial<Record<SwapContractNames, SystemConfigItem>>
```

scripts/common/config-item.ts
```diff
@@ -1,26 +1,98 @@
...
+type CoreRecord = Record<CoreContractNamesWithoutSwap, SystemConfigItem>
+type AaveV2ActionsRecord = PartialRecord<AaveV2ContractNames, SystemConfigItem>
```
peculiarity:
```suggestion
type AaveV2ActionsRecord = Partial<Record<AaveV2ContractNames, SystemConfigItem>>
```

scripts/common/config-item.ts
```diff
@@ -1,26 +1,98 @@
+import { PartialRecord } from '@helpers/types/common'
```
peculiarity:
```suggestion
```

scripts/common/deploy-system.ts
```diff
@@ -1,4 +1,9 @@
+import { Network } from '@helpers/network'
+import { ServiceRegistry as ServiceRegistryClass } from '@helpers/serviceRegistry'
+import { PartialRecord } from '@helpers/types/common'
```
peculiarity:
```suggestion
```

scripts/common/deploy-system.ts
```diff
@@ -28,10 +33,29 @@ import {
...
-import { Network } from './types'
 import { removeVersion } from './utils'
 
+export type ContractProps = {
+  contract: Contract
+  config: ConfigItem | Record<string, unknown>
+  hash: string
+}
+
+export type SystemTemplate20 = PartialRecord<DeployedSystemContractNames, ContractProps>
```
peculiarity:
```suggestion
export type SystemTemplate20 = Partial<Record<DeployedSystemContractNames, ContractProps>>
```

packages/oasis-actions/src/strategies/aave/close/close.ts
```diff
@@ -288,42 +304,74 @@ async function buildOperation(
...
-    receiveAtLeast: swapData.minToTokenAmount,
-    proxy: dependencies.proxy,
-    collectFeeFrom: swapData.collectFeeFrom,
-    collateralTokenAddress,
-    collateralIsEth: args.collateralToken.symbol === 'ETH',
-    debtTokenAddress,
-    debtTokenIsEth: args.debtToken.symbol === 'ETH',
-    isDPMProxy: dependencies.isDPMProxy,
-    shouldCloseToCollateral: args.shouldCloseToCollateral || false,
+  const lockedCollateralAmountInWei = args.shouldCloseToCollateral
```
peculiarity: If I understand correctly that is the amount that is going to be swapped. I think the naming is very misleading.  The approach is good but the name could be  something like `swappableCollateralAmount` or `collateralAmountToBeSwapped` or  `swapAmount` or something like that.  The locked collateral amount is what we currently have in the position. After close it will be always 0 no matter what you are closing to. You simply won't have any collateral locked, right ?

scripts/common/config-item.ts
```diff
@@ -1,26 +1,98 @@
...
+  | 'WrapEth'
+  | 'UnwrapEth'
+  | 'TakeFlashloan'
+  | 'ReturnFunds'
+  | 'PositionCreated'
+  | 'AaveV3Borrow'
+  | 'AaveV3Deposit'
+  | 'AaveV3Withdraw'
+  | 'AaveV3Payback'
+  | 'AaveV3SetEMode'
```
peculiarity:
```suggestion
export type AAVEV3ActionNames = 'AaveV3Borrow'
  | 'AaveV3Deposit'
  | 'AaveV3Withdraw'
  | 'AaveV3Payback'
  | 'AaveV3SetEMode'
  
export type CoreActionNames = | 'SwapAction'
  | 'PullToken'
  | 'SendToken'
  | 'SetApproval'
  | 'WrapEth'
  | 'UnwrapEth'
  | 'TakeFlashloan'
  | 'ReturnFunds'
  | 'PositionCreated'

```

scripts/common/config-item.ts
```diff
@@ -1,26 +1,98 @@
...
+type AaveProtocolContractNamesV3 = 'AaveOracle' | 'Pool' | 'AaveProtocolDataProvider'
+type AaveProtocolContractNames = AaveProtocolContractNamesV2 | AaveProtocolContractNamesV3
+
+export type AllowedContractNames =
+  | CoreContractNames
+  | ActionContractNames
+  | CommonContractNames
+  | AaveProtocolContractNames
+
+export type LocalSystemContractNames = 'DsProxyRegistry'
```
peculiarity: Why do we need this per se?

scripts/common/deploy-system.ts
peculiarity: Also let's remove that 20 from the type.

scripts/deployment20/test-configs/mainnet.conf.ts
```diff
@@ -43,6 +43,30 @@ export const config: Config = {
...
+        constructorArgs: ['address:DSProxyFactory'],
+      },
+      DSGuardFactory: {
+        name: 'DSGuardFactory',
+        deploy: true,
+        address: '',
+        serviceRegistryName: CONTRACT_NAMES.common.DS_GUARD_FACTORY,
+        history: [],
+        constructorArgs: [],
+      },
```
peculiarity:
```suggestion
      
```

These shouldn't be here. They are already deployed there.

scripts/deployment20/test-configs/opt-mainnet.conf.ts
```diff
@@ -88,6 +90,14 @@ export const config = {
           'address:ServiceRegistry',
         ],
       },
+      ChainLogView: {
```
peculiarity: Let's rename the whole file to `optimism.conf` again. Also we don't need that/ You don't maker's `chainlog` on optimism. Also no idea where you go that  address from ( the one for the constructor )

test/test-utils/one-inch-version.ts
```diff
@@ -0,0 +1,19 @@
+import { Network } from '@helpers/network'
+
+type OneInchVersion = 'v4.0' | 'v5.0'
```
peculiarity: Check `OneInchCall.ts` and see what could be changed ther based on this. I see that this type can be applied there. Actually you can put that logic in that file instead of creating this one. We start having so many files with same domain logic across the app.

peculiarity (changes requested):

scripts/common/config-item.ts
zerotucks: We don't. Relic from before merging your stuff

helpers/aave/buildGetTokenByImpersonateFunction.ts
```diff
@@ -4,37 +4,65 @@ import BigNumber from 'bignumber.js'
...
+      whale: EMPTY_ADDRESS,
+    },
   },
 }
 
-// TODO: Needs to work with different networks eventually
 export function buildGetTokenByImpersonateFunction(
   config: RuntimeConfig,
   hre: HardhatRuntimeEnvironment,
+  network: Network.MAINNET | Network.OPTIMISM,
```
peculiarity, line 59: Wasn't there a `Network` type? If we add `Arbitrum` we will have to go all over these places and add it as well. Probably would be nice to have all these places where you depend on `Network.MAINNET | Network.OPTIMISM` be replaced with something more generic that will be extended in the future.

scripts/common/config-item.ts
```diff
@@ -2,25 +2,97 @@ import { ContractNames } from '@oasisdex/oasis-actions/src'
...
+  | 'DSGuardFactory'
+  | 'AccountGuard'
+  | 'AccountFactory'
+
+export type CoreContractNames =
+  | CoreContractNamesWithoutSwap
+  | SwapContractNames
+  | MainnetOnlyCoreContractNames
+export type AaveV2ContractNames = 'AaveBorrow' | 'AaveDeposit' | 'AaveWithdraw' | 'AavePayback'
+export type ActionContractNamesWithoutAaveV2 =
```
peculiarity: Why do we still aave ations mixed into the core actions type?

peculiarity: This is how I imagine it:

```
export type CoreContracts =
  | 'ServiceRegistry'
  | 'OperationExecutor'
  | 'OperationStorage'
  | 'OperationsRegistry'
  | 'DSProxyFactory'
  | 'DSProxyRegistry'
  | 'DSGuardFactory'
  | 'AccountGuard'
  | 'AccountFactory'
  | 'ChainLogView'
  | 'Swap'
  | 'uSwap'

export type AaveV2Actions =
  | 'AaveBorrow'
  | 'AaveDeposit'
  | 'AaveWithdraw'
  | `AavePayback`
  | 'AavePayback'

export type AaveV3Actions =
  | `AaveV3Borrow`
  | `AaveV3Deposit`
  | `AaveV3Withdraw`
  | `AaveV3Payback`
  | `AaveV3SetEMode`
  
  export type CoreActions =
  | 'SwapAction'
  | 'PullToken'
  | 'SendToken'
  | 'SetApproval'
  | 'WrapEth'
  | 'UnwrapEth'
  | 'TakeFlashloan'
  | 'ReturnFunds'
  | 'PositionCreated'
  
  export type Actions = CoreActions | AaveV2Actions | AaveV3Actions
  ```

File: scripts/common/config-item.ts [hunk 18]
```diff
@@ -2,25 +2,97 @@ import { ContractNames } from '@oasisdex/oasis-actions/src'
...
+  | 'DSProxyRegistry'
+  | 'DSGuardFactory'
+  | 'AccountGuard'
+  | 'AccountFactory'
+
+export type CoreContractNames =
+  | CoreContractNamesWithoutSwap
+  | SwapContractNames
+  | MainnetOnlyCoreContractNames
+export type AaveV2ContractNames = 'AaveBorrow' | 'AaveDeposit' | 'AaveWithdraw' | 'AavePayback'
```
peculiarity: btw you missed `AavePayback` here

scripts/common/config-item.ts
```diff
@@ -2,25 +2,97 @@ import { ContractNames } from '@oasisdex/oasis-actions/src'
...
+  | 'ChainlinkEthUsdPriceFeed'
+type AaveProtocolContractNamesV2 =
+  | 'PriceOracle'
+  | 'LendingPool'
+  | 'ProtocolDataProvider'
+  | 'WETHGateway'
+type AaveProtocolContractNamesV3 = 'AaveOracle' | 'Pool' | 'AaveProtocolDataProvider'
+type AaveProtocolContractNames = AaveProtocolContractNamesV2 | AaveProtocolContractNamesV3
+
+export type AllowedContractNames =
```
peculiarity: Why is it called allowed?

zerotucks: I've tried to make these types clearer based on your feedback. Lmk what you think

peculiarity (approved):

PR #250, opened 2023-03-31
Calculate liquidationPrice in AjnaPosition

piotrkonowrocki:
Changes:
- added `pendingInflator` to `ajnaPool` interface,
- calculated `liquidationPrice` in `AjnaPosition`.

### DIFF
--- a/packages/oasis-actions/src/types/ajna/AjnaPosition.ts
+++ b/packages/oasis-actions/src/types/ajna/AjnaPosition.ts
@@ -47,12 +47,7 @@ export class AjnaPosition implements IAjnaPosition {
 
   get liquidationPrice() {
     const liquidationPrice = this.pool.mostOptimisticMatchingPrice
-      .times(
-        this.debtAmount
-          .div(this.pool.pendingInflator)
-          .times(this.pool.pendingInflator)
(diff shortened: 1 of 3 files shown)


packages/oasis-actions/src/types/ajna/AjnaPosition.ts
```diff
@@ -45,7 +45,16 @@ export class AjnaPosition implements IAjnaPosition {
   ) {}
 
   get liquidationPrice() {
-    return ZERO
+    const liquidationPrice = this.pool.mostOptimisticMatchingPrice
+      .times(
+        this.debtAmount
+          .div(this.pool.pendingInflator)
+          .times(this.pool.pendingInflator)
```
piekczyk: Just to double check, Is this correct? I mean `div` by pending inflator and then `times`?

piotrkonowrocki: Good question, but according to formulas in Notion it is:
`t0debt * pendingInflator`
and `t0debt` is `debtAmount / pendingInflator`

piotrkonowrocki: Removed.

piekczyk (approved):

PR #251, opened 2023-04-04
HH Task > Open AAVE V3 Multiply Positions on Optimism

zerotucks:
**Changes**
- Removed `createPosition` task which was well out of date
- Updated `createMultiplePosition(s)` task to work across networks and also to work only with V3 positions we know are cross L1/L2 compatible
- Merged with mono changes

**To Test**
1. `cd packages/dma-contracts`
1. `yarn dev:optimism`
2. `yarn deploy:dev`
3. `yarn hardhat createMultiplyPositions --serviceregistry <sr-address-from-previous-step> --accountfactory <account-factory-address-from-previous-step> --network local`


scripts/common/impersonate.ts
```diff
@@ -0,0 +1,15 @@
+import hre from 'hardhat'
+
+export const impersonateAccount = async (account: string) => {
```
jakubswierczek: is it needed here as we have it in `hardhat.utils.ts`?

jakubswierczek (approved):

PR #252, opened 2023-04-13
Mono | Use `dma` deployment config addresses as source of truth for `addresses` package

zerotucks:
Note: Please read **Relevant Files** section before reviewing. Most files are simple import changes and can be ignored.

**Changes**
* Remove old address json files
* Convert @oasisdex/addresses into a package /src directory builds to /lib
* Updated all ADDRESSES imports to rely on source code not built version (@oasisdex/addresses/src) // many files
* Cleaned up network configs so they conform to SystemConfig type

**Relevant Files**
* packages/addresses/src/index.ts
* packages/dma-contracts/test/addresses/mainnet.ts // shows example of ADDRESSES record being used.
* packages/dma-deployments/configs/goerli.conf.ts // shows how configs have been typed appropriately
* packages/dma-deployments/configs/mainnet.conf.ts // as above

PR #254, opened 2023-04-13
FIX - missing `lerna` dependency

halaprix:


zerotucks:
This may no longer be needed based on this PR being merged https://github.com/OasisDEX/oasis-earn-sc/pull/252

PR #256, opened 2023-04-14
Ajna earn empty position handling

piekczyk:
[Ajna earn empty position handling](https://app.shortcut.com/oazo-apps/story/8734/bug-ajna-earn-empty-position)

- added handling for reopening earn position
- added handling for empty position
- handled cases for nft / no nft positions
- adjusted max withdraw validation so it doesn't return error message due to rounding from UI

### DIFF
--- a/packages/oasis-actions/src/strategies/ajna/earn/depositAndAdjust.ts
+++ b/packages/oasis-actions/src/strategies/ajna/earn/depositAndAdjust.ts
@@ -19,7 +19,9 @@ export async function depositAndAdjust(
     args.position.pool.quoteToken.toLowerCase() === dependencies.WETH.toLowerCase()
   const isPositionStaked = args.position.stakedNftId !== null
   const isDepositing = args.quoteAmount.gt(ZERO)
-  const isAdjusting = !args.price.eq(args.position.price)
+  const isAdjusting = !args.price.eq(args.position.price) && args.position.price.gt(ZERO)
+  const isReopening = args.position.price.isZero()
+  const shouldStakeNft = args.isStakingNft
 
   const ajnaProxyActions = new ethers.Contract(
     dependencies.ajnaProxyActions,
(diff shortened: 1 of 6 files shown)


packages/oasis-actions/src/strategies/ajna/earn/depositAndAdjust.ts
```diff
@@ -108,6 +112,26 @@ export async function depositAndAdjust(
     targetPosition = args.position.deposit(args.quoteAmount)
   }
 
+  if (!isPositionStaked && isDepositing && isReopening && shouldStakeNft) {
```
piotrkonowrocki: Am I missing something or the only difference between those two conditions is second param in `encodeFunctionData`?
Then it could be shorter:
```ts
  if (!isPositionStaked && isDepositing && isReopening) {
    data = ajnaProxyActions.interface.encodeFunctionData(
      shouldStakeNft ? 'supplyQuoteMintNftAndStake' : 'supplyQuote',
        [
          args.poolAddress,
          ethers.utils.parseUnits(args.quoteAmount.toString(), args.quoteTokenPrecision).toString(),
          args.price.shiftedBy(18).toString(),
        ]
      )
    targetPosition = args.position.reopen(args.quoteAmount, priceToIndex)
  }
```

packages/oasis-actions/src/strategies/ajna/earn/withdrawAndAdjust.ts
```diff
@@ -106,6 +107,26 @@ export async function withdrawAndAdjust(
     targetPosition = args.position.withdraw(args.quoteAmount)
   }
 
+  if (!isPositionStaked && isWithdrawingAll) {
```
piotrkonowrocki, line 110: Same as in above.

piotrkonowrocki (changes requested):

packages/oasis-actions/src/types/ajna/AjnaEarnPosition.ts
```diff
@@ -154,4 +155,30 @@ export class AjnaEarnPosition implements IAjnaEarn {
...
+      this.pool,
+      this.owner,
+      quoteTokenAmount,
+      priceIndex,
+      this.stakedNftId,
+      this.collateralPrice,
+      this.quotePrice,
+      this.rewards,
+    )
+  }
```
paszkowskiDamian, line 170: is reopen needed? Can we just deposit into empty position?

piekczyk: yes, I tried with deposit initially and I would need to overwrite princeIndex as well since the one from empty position is 0

packages/oasis-actions/src/strategies/ajna/earn/depositAndAdjust.ts
piekczyk: good catch, fixed

packages/oasis-actions/src/strategies/ajna/earn/withdrawAndAdjust.ts
piekczyk: here it's not applicable since value interface is different

piotrkonowrocki (approved):

PR #258, opened 2023-04-17
Ajna actions2

paszkowskiDamian:


contracts/actions/ajna/DepositBorrow.sol
```diff
@@ -0,0 +1,39 @@
...
+   * @param data Encoded calldata that conforms to the BorrowData struct
+   */
+  function execute(bytes calldata data, uint8[] memory) external payable override {
+    DepositBorrowData memory args = parseInputs(data);
+    IAjnaPool pool = IAjnaPool(args.pool);
+    IAjnaPoolUtilsInfo poolUtilsInfo = IAjnaPoolUtilsInfo(registry.getRegisteredService(AJNA_POOL_UTILS_INFO));
+
+    uint256 index = poolUtilsInfo.priceToIndex(args.price);
+
+    pool.drawDebt(address(this), args.borrowAmount * pool.quoteTokenScale(), index, args.borrowAmount * pool.collateralScale());
```
halaprix, line 31: `args.borrowAmount` is used twice

contracts/actions/ajna/RepayWithdraw.sol
```diff
@@ -0,0 +1,39 @@
...
+   */
+  function execute(bytes calldata data, uint8[] memory) external payable override {
+    RepayWithdrawData memory args = parseInputs(data);
+    IAjnaPool pool = IAjnaPool(args.pool);
+    IAjnaPoolUtilsInfo poolUtilsInfo = IAjnaPoolUtilsInfo(registry.getRegisteredService(AJNA_POOL_UTILS_INFO));
+
+    uint256 index = poolUtilsInfo.priceToIndex(args.price);
+
+    pool.repayDebt(address(this), args.repayAmount * pool.quoteTokenScale(), args.withdrawAmount * pool.collateralScale(), address(this), index);
+    store().write(bytes32(args.repayAmount));
```
halaprix, line 32: the order is different thatn in `DepositBorrow` is it intended?

- 1st: debt 
- 2nd: collateral

halaprix (changes requested):
two comments

contracts/actions/ajna/RepayWithdraw.sol
paszkowskiDamian: it was supposed to reflect the order of actual operations, first you need to repay debt and you can withdraw, also this is exactly the same as in action name.

paszkowskiDamian:
@halaprix https://github.com/OasisDEX/oasis-earn-sc/pull/280 requested changes were applied here.

PR #263, opened 2023-04-18
Adjusted getFeeWhenBelowLup and add origiantion fee cals

piekczyk:
- adjusted `getFeeWhenBelowLup` since it should be in quote token unit
- added `originationFee` calculations

### DIFF
--- a/packages/oasis-actions/src/helpers/ajna/index.ts
+++ b/packages/oasis-actions/src/helpers/ajna/index.ts
@@ -112,3 +112,14 @@ export const calculateAjnaApyPerDays = (amount: BigNumber, apy: BigNumber, days:
     .times(new BigNumber(Math.E ** apy.times(days).div(365).toNumber()))
     .minus(amount)
     .div(amount)
+
+// The origination fee is calculated as the greatest of the current annualized
+// borrower interest rate divided by 52 (one week of interest) or 5 bps multiplied by the loan’s new
+// debt.
+export const getAjnaBorrowOriginationFee = ({
+  interestRate,
+  quoteAmount,
+}: {
+  interestRate: BigNumber
+  quoteAmount: BigNumber
+}) => BigNumber.max(interestRate.div(52), new BigNumber(0.0005)).times(quoteAmount)
--- a/packages/oasis-actions/src/helpers/ajna/index.ts
+++ b/packages/oasis-actions/src/helpers/ajna/index.ts
@@ -122,4 +122,9 @@ export const getAjnaBorrowOriginationFee = ({
 }: {
   interestRate: BigNumber
   quoteAmount: BigNumber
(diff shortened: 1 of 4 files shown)


packages/oasis-actions/src/helpers/ajna/index.ts
```diff
@@ -112,3 +112,14 @@ export const calculateAjnaApyPerDays = (amount: BigNumber, apy: BigNumber, days:
...
+// The origination fee is calculated as the greatest of the current annualized
+// borrower interest rate divided by 52 (one week of interest) or 5 bps multiplied by the loan’s new
+// debt.
+export const getAjnaBorrowOriginationFee = ({
+  interestRate,
+  quoteAmount,
+}: {
+  interestRate: BigNumber
+  quoteAmount: BigNumber
+}) => BigNumber.max(interestRate.div(52), new BigNumber(0.0005)).times(quoteAmount)
```
paszkowskiDamian: Can you name the magic numbers? 🙏

packages/oasis-actions/src/types/ajna/AjnaPosition.ts
```diff
@@ -94,6 +96,13 @@ export class AjnaPosition implements IAjnaPosition {
     return new RiskRatio(normalizeValue(loanToValue), RiskRatio.TYPE.LTV)
   }
 
+  originationFee(quoteAmount: BigNumber) {
+    return getAjnaBorrowOriginationFee({
+      interestRate: this.pool.interestRate,
+      quoteAmount,
+    })
+  }
+
```
paszkowskiDamian, line 105: Can you move origininationFee to simulation? 

I would add new fields to simulation called fees which will be an array of fees, with amount, token and maybe name 

What do you think?

packages/oasis-actions/src/helpers/ajna/index.ts
piekczyk: I added description above this method to clarify what 52 and 0.0005 means. I thought about extracting it as const but I doubt if we ever reuse it since these seems to be method specific.

packages/oasis-actions/src/types/ajna/AjnaPosition.ts
piekczyk: I'm using this method in borrow as well so it must be defined on constructor level.

packages/oasis-actions/src/helpers/ajna/index.ts
piotrkonowrocki: I agree about naming the number. Even if it's not going to be reused, it's better to take this number and write it as a const at the beginning of the file, without exporting it. That'll make it easier to read.

piotrkonowrocki (approved):

PR #265, opened 2023-04-19
@oasisdex/addresses bundling

marcinciarka:
Provides a way to build/bundle the package (for easy use as npm package).


zerotucks (approved):
Have contributed. That's my review. See esm / cjs build output changes

PR #270, opened 2023-04-20
FEAT - add `ajna-contracts`

halaprix:


packages/dma-contracts/README.md
```diff
@@ -0,0 +1,7 @@
+# @oasisdex/dma-contracts
+
+## RUN BEFORE USING DMA-CONTRACTS
+You need to create a symlink, to include `ajna-contracts` in `dma-contracts` for proper solc compilation.
+```bash
+/oasis-earn-sc/packages/dma-contracts$ ln -s ~/absolute_path_to_the_repository/oasis-earn-sc/packages/ajna-contracts/contracts/ajna ./contracts/
+```
```
paszkowskiDamian, line 7: Isn't lerna supposed to do it?

packages/abis/system/contracts/interfaces/dpm/IAccountGuard.sol/IAccountGuard.json
```diff
@@ -1,4 +1,98 @@
 [
```
paszkowskiDamian, line 1: Shouldn't we use abis and contracts already deified? Or this is the issue with hardhat not being able to find Contract Factories?

packages/dma-contracts/README.md
halaprix: it's an additional symlink to make hh compile contracts from `ajna-contracts` while running compile in `dma-contracts` . temporary solution, and after ajna release we will get rid of this package or at least not use it in `dma-contracts`

packages/abis/system/contracts/interfaces/dpm/IAccountGuard.sol/IAccountGuard.json
halaprix: Let me check that tomorrow. For sure I've expanded the interfaces for guard and account, but the same contract sources are used.

paszkowskiDamian (approved):

PR #301, opened 2023-05-15
Library exports adjustment

piekczyk:
- fixed issue with duplicated types in dma-library 
- partially unified `bignumber.js` and `ethers` packages versions across packages within monorerpo  and borrow repo


zerotucks (approved):
Looks good to me. 

Caveat: we'll need to address aave types as currently aave operations are type "any". (https://github.com/OasisDEX/oasis-earn-sc/blob/8b883a69e22897deffb17b9dd97670df0093a5a7/packages/dma-library/src/operations/index.ts) Needs to be done prior to integrating in OB re L2 AAVE - unless library was published end of last week. cc: @marcinciarka @jakubswierczek (one of us will need to do before publishing library again re AAVE)

PR #302, opened 2023-05-15
FIX - mono deployments

zerotucks:
**Problem**
Deploy scripts in dma-deployments (now renamed to deploy-configurations) did not have access to contracts and compiled artifacts. The deployment configs (that contain contract addresses) are used as a single source of truth for system addresses and are consumed by /addresses package `@oasisdex/addresses` so need to remain at the root of our package dependency graph to stop us from running into cyclical dependency issues.

**Solution**
The deploy scripts themselves can be safely moved back into dma-contracts where they'll have direct access to contracts and build artifacts when deploying. Those deployment scripts are also specific to `DMA` so this makes sense.

**Longer term**
`deploy-configurations` can used more generally for any systems we want to deploy EG Automation contracts and so on.


marcinciarka (approved):
LGTM

PR #307, opened 2023-05-17
chore: add events to `AjnaProxyActions`

halaprix:
- add events to AjnaProxyActions as per : https://docs.google.com/spreadsheets/d/1lrjJ3pAkZFTViNJiR1sJ2Iphyf_fGnlGwvZIZ-AA3eY/edit#gid=793926827

- event ProxyActionsOperation so it can be indexed separately to Operaion from DMA

### DIFF
--- a/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
+++ b/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
@@ -42,6 +42,13 @@ contract AjnaProxyActions {
         GUARD = _GUARD;
     }
 
+    /**
+     * @dev Emitted once an Operation has completed execution
+     * @param name The address initiating the deposit
+     **/
+    event ProxyActionsOperation(bytes32 indexed name);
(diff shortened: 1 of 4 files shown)


paszkowskiDamian (approved):
As I mentioned it seems like a lot of different Operations but maybe this is the way. Let's proceed 👍

packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
```diff
@@ -42,6 +42,20 @@ contract AjnaProxyActions {
         GUARD = _GUARD;
     }
 
+    /**
+     * @dev Emitted once an Operation has completed execution
+     * @param name Name of the operation
+     **/
+    event ProxyActionsOperation(bytes32 indexed name);
```
paszkowskiDamian, line 49: I see that you are emitting strings where event has `bytes32`. Did you verify if events will have decodable names or hashes? For simplicity you might need to encode those strings to bytes32 before emitting event.

PR #317, opened 2023-05-19
[@oasisdex/addresses@0.0.15] Aave v3 arbitrum config

marcinciarka:
This is including only the basic stuff (tokens and contract addresses) regarding the Aave V3 product cards.

### DIFF
--- a/packages/deploy-configurations/addresses/index.ts
+++ b/packages/deploy-configurations/addresses/index.ts
@@ -1,3 +1,10 @@
+import {
+  arbitrumConfig,
+  goerliConfig,
+  mainnetConfig,
+  optimismConfig,
+} from '@deploy-configurations/configs'
+import { ADDRESS_ZERO as zeroAddress } from '@deploy-configurations/constants'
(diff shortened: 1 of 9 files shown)


zerotucks (approved):
LGTM.

packages/deploy-configurations/addresses/index.ts
```diff
@@ -12,11 +12,12 @@ import {
   MakerProtocol,
   MakerProtocolJoins,
   MakerProtocolPips,
+  SystemConfig,
   SystemKeys,
 } from '@deploy-configurations/types/deployment-config'
 import { Network } from '@deploy-configurations/types/network'
 
-import { goerliConfig, mainnetConfig, optimismConfig } from '../configs'
+import { arbitrumConfig, goerliConfig, mainnetConfig, optimismConfig } from '../configs'
```
zerotucks: Non blocking comment. But, trying to move to non relative paths. Want to enforce this in linting rules shortly.

marcinciarka: Fixed that.

PR #333, opened 2023-05-31
Ajna adjust risk down operation

piekczyk:
- added ajna adjust risk down operation

### DIFF
--- a/packages/dma-library/src/operations/ajna/adjust-risk-down.ts
+++ b/packages/dma-library/src/operations/ajna/adjust-risk-down.ts
@@ -64,16 +64,18 @@ export const adjustRiskDown: AjnaAdjustRiskDownOperation = async ({
   const paybackWithdraw = actions.ajna.ajnaPaybackWithdraw(
     {
       pool: addresses.pool,
-      paybackAmount: ZERO,
       withdrawAmount: collateral.withdrawal.amount,
+      paybackAmount: ZERO,
       price,
     },
-    [0, 1, 0, 0, 0, 0],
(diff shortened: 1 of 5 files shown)


packages/dma-library/src/operations/ajna/adjust-risk-down.ts
```diff
@@ -0,0 +1,105 @@
...
+  proxy,
+  addresses,
+  price,
+}) => {
+  const setDebtTokenApprovalOnPool = actions.common.setApproval({
+    asset: debt.address,
+    delegate: addresses.pool,
+    amount: flashloan.amount,
+    sumAmounts: false,
+  })
```
paszkowskiDamian: approval need to be set up after swap. The amount is unknown before swap.

piekczyk: fixed

packages/dma-library/src/operations/ajna/adjust-risk-down.ts
```diff
@@ -0,0 +1,110 @@
...
+    asset: collateral.isEth ? addresses.ETH : collateral.address,
+  })
+
+  const flashloanCalls = [
+    setDebtTokenApprovalOnPool,
+    paybackWithdraw,
+    swapCollateralTokensForDebtTokens,
+    unwrapEth,
+    returnDebtFunds,
+    returnCollateralFunds,
```
paszkowskiDamian: As we discussed it might be necessary to move returns after FL. Honestly I don't remember how it worked before

paszkowskiDamian (approved):

PR #335, opened 2023-05-31
implement getPoolData

paszkowskiDamian:
Implementation of helper function that strategy needs as a dependency.


packages/ajna-contracts/scripts/prepare-env.ts
```diff
@@ -21,6 +21,51 @@ const plotly =
...
+  quoteTokens: BigNumber;
+  collateral: BigNumber;
+  bucketLPs: BigNumber;
+}
+interface AjnaPool {
+  poolAddress: string;
+  quoteToken: string;
+  collateralToken: string;
+
+  //@deprecated use lowestUtilizedPrice
```
zerotucks, line 37: Minor point but /** @deprecated as it's picked up by IDE

zerotucks (approved):

PR #338, opened 2023-06-01
Ajna close operations

piekczyk:
https://app.shortcut.com/oazo-apps/story/8670/common-library-complex-operations-close

- addded ajna close to quote and collateral operations


zerotucks (approved):
Approving. If there are any minor issues with operations we can resolve them at the point where we implement the strategy

PR #345, opened 2023-06-08
Clarify Precision + Precision Aware Amount class

zerotucks:
**Changes**
1. Create immutable BigNumber wrapper that is precision aware called `Amount`
2. Refactor Aave Open Multiply and Ajna Open Multiply to use this class in place of amountFromWei and amountToWei utilities
3. Refactor domain logic to use Amount class for all precision handling
4. Begin updating variable names in domain logic and two open multiply strategies to use the $ (for max token precision) $$ (for normalised values) convention. As per the readme in the domain package.


packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+type AmountArgs = {
+  amount: BigNumber
+  precision?: {
+    /** none: 1USDC = 1, max: 1USDC = 1e6, normalized: 1USDC = 1e18 */
+    mode?: PrecisionMode
+    tokenMaxDecimals?: number
+  }
+}
+
+export class Amount {
```
jakubswierczek, line 27: I don't think a wrapper for blockchain technology is domain-specific; it should be in a separate package, independent of the domain.

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+ * - (max precision for the token)
+ *
+ * - normalized                                         EG 1 USDC = 1e18
+ * - (precision set 18 decimals for to standardise amounts for maths reasons)
+ */
+type PrecisionMode = 'none' | 'tokenMax' | 'normalized'
+type PrecisionMap = Record<PrecisionMode, BigNumber>
+
+type AmountArgs = {
+  amount: BigNumber
```
jakubswierczek, line 19: that should be `string` or `number`, maybe `BigInt`. We should hide any dependencies.

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  private _verifyAmountsAreCompatible(otherAmount: Amount) {
+    const thisTokenMaxDecimals = this._getTokenMaxPrecisionAsBigNumber()
+    const otherTokenMaxDecimals = otherAmount._getTokenMaxPrecisionAsBigNumber()
+
+    const thisPrecisionMode = this.getCurrentPrecisionMode()
+    const otherPrecisionMode = otherAmount.getCurrentPrecisionMode()
+
+    if (thisPrecisionMode === 'normalized' && otherPrecisionMode === 'normalized') {
+      return true // All good
```
jakubswierczek, line 171: no need to return `true`

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+    return new Amount({
+      amount: newAmount,
+      precision: {
+        mode: currentPrecisionMode,
+        tokenMaxDecimals: this.getTokenMaxDecimals(),
+      },
+    })
+  }
+
+  private _verifyAmountsAreCompatible(otherAmount: Amount) {
```
jakubswierczek, line 163: make it static and pass two amounts

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+    if (thisTokenMaxDecimals.eq(otherTokenMaxDecimals)) {
+      return true // All good
+    }
+
+    throw new Error(
+      `Amounts are not compatible: ${thisTokenMaxDecimals.toString()} !== ${otherTokenMaxDecimals.toString()}`,
+    )
+  }
+
+  private _getTokenMaxPrecisionAsBigNumber(): BigNumber {
```
jakubswierczek, line 183: decimals could be just a number

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+  private currentPrecisionScale: BigNumber
+  private readonly tokenMaxDecimals: number
+  private amount: BigNumber
+  private readonly precisionMap: PrecisionMap
+  private currentPrecisionMode: 'none' | 'tokenMax' | 'normalized' = 'none'
+
+  constructor(args: AmountArgs) {
+    const initialAmount = args.amount || ZERO
+    const initialPrecisionMode = args?.precision?.mode || 'none'
+    this.tokenMaxDecimals = args?.precision?.tokenMaxDecimals || TYPICAL_PRECISION
```
jakubswierczek, line 37: default should be `undefined`

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  plus(otherAmount: Amount | BigNumber): Amount {
+    if (otherAmount instanceof Amount) {
+      return this._doMaths(otherAmount, 'plus')
+    }
+    return this._plusByBigNumber(otherAmount)
+  }
+
+  minus(otherAmount: Amount | BigNumber): Amount {
```
jakubswierczek, line 55: like before, let's to the `Amount` or `string`
Or make an Interface `Amountish`

packages/domain/src/amount.ts
jakubswierczek, line 27: It should have methods for `toString`, `toHexString` 

Some wishes:
- `toString` has a format -> percentage, normal etc
- `console.log(amount)` should print amount in string

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+export class Amount {
+  private currentPrecisionScale: BigNumber
+  private readonly tokenMaxDecimals: number
+  private amount: BigNumber
+  private readonly precisionMap: PrecisionMap
+  private currentPrecisionMode: 'none' | 'tokenMax' | 'normalized' = 'none'
+
+  constructor(args: AmountArgs) {
```
jakubswierczek, line 34: let's change approach. Make constructor to have two args. One is a value (string, number, BigInt), second is optional (with question mark) and this is an options. 
```
constructor(value: Valuish, options?: AmountOptions) 
```

packages/domain/src/amount.ts
```diff
@@ -0,0 +1,186 @@
...
+    const initialAmount = args.amount || ZERO
+    const initialPrecisionMode = args?.precision?.mode || 'none'
+    this.tokenMaxDecimals = args?.precision?.tokenMaxDecimals || TYPICAL_PRECISION
+    this.precisionMap = {
+      none: ONE,
+      tokenMax: new BigNumber(TEN.pow(this.tokenMaxDecimals)),
```
jakubswierczek, line 40: I'm not a fan of the name `tokenMax`. A big number is usually used to represent a large quantity, not a balance. It could be a percentage. It's like creating a wrapper for a float number and passing a currency symbol as a parameter.

jakubswierczek (changes requested):

zerotucks:
@jakubswierczek Thank you for review. 😀 Perhaps we can pick up in the future.
@paszkowskiDamian Closing for now after catchup 👍 🙏

PR #349, opened 2023-06-14
Ajna Multiply Close Position

zerotucks:
**Changes**
* Close to quote and collateral logic in ajna/multiply/close strategy file
* Test file using new ajna env to test closing ETH/USDC position both to collateral and quote

**To test**
`cd to dma/contracts` for ease
`yarn hardhat test test/e2e/strategies/ajna/close.test.ts`


paszkowskiDamian:
I think this one needs to be refactored so it does not include the new Amount class (I think the new approach needs broader discussion with team)

zerotucks:
Closed in favour of https://github.com/OasisDEX/oasis-earn-sc/pull/358

PR #358, opened 2023-06-26
Ajna Close Multiply Strategy

zerotucks:
Replication of this PR -> https://github.com/OasisDEX/oasis-earn-sc/pull/349

With precision aware Amount class and precision notation stripped out as per closed https://github.com/OasisDEX/oasis-earn-sc/pull/345


packages/dma-contracts/test/e2e/strategies/ajna/close.test.ts
```diff
@@ -0,0 +1,306 @@
...
+          {
+            config: env.config,
+            isFormatted: true,
+          },
+        )
+
+        expect.toBeEqual(proxyDebtBalance, ZERO)
+        expect.toBeEqual(proxyCollateralBalance, ZERO)
+      })
+      it(`Should have collected a fee for ${variant}`, async () => {
```
paszkowskiDamian, line 123: We should also test whether the user got the collateral afterwards.

packages/dma-library/src/operations/ajna/close-to-collateral.ts
```diff
@@ -0,0 +1,113 @@
...
+    amount: new BigNumber(MAX_UINT),
+  })
+
+  unwrapEth.skipped = !debt.isEth && !collateral.isEth
+
+  const sendQuoteTokenToOpExecutor = actions.common.sendToken({
+    asset: debt.address,
+    to: addresses.operationExecutor,
+    amount: flashloan.amount.plus(BALANCER_FEE.div(FEE_BASE).times(flashloan.amount)),
+  })
```
paszkowskiDamian: in case when we were flashfloaning WETH, shouldn't we send it back to OP before unwrapping it?

zerotucks: Yes absolutely. Great point.

packages/dma-contracts/test/e2e/strategies/ajna/close.test.ts
zerotucks: I'll implement one. Thanks dude 💯

PR #366, opened 2023-07-04
Kk/final-apa-changes

halaprix:
- `repayWithdraw` and `depositAndDraw` had two ajna pool contract calls. That caused two events being emitted and slightly higher gas cost.
- only the event emission logic is in place 
- added `AjnaRepayAndClose` event

### DIFF
--- a/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
+++ b/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
@@ -264,7 +264,7 @@ contract AjnaProxyActions is IAjnaProxyActions {
      *  @param  pool           Pool address
      *  @param  amount         Amount of debt to repay
      */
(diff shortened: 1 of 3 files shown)


packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
```diff
@@ -285,12 +285,19 @@ contract AjnaProxyActions is IAjnaProxyActions {
         uint256 collateralAmount,
         uint256 price
     ) public payable {
-        if (collateralAmount > 0) {
-            depositCollateral(pool, collateralAmount, price);
-        }
-        if (debtAmount > 0) {
-            drawDebt(pool, debtAmount, price);
```
paszkowskiDamian, line 292: Shouldn't you remove unused functions`depositCollateral` and `drawDebt`?

paszkowskiDamian (approved):

PR #389, opened 2023-07-17
PnL method on AjnaPosition

zerotucks:
## Ticket URL

Please provide a link to ticket:
https://app.shortcut.com/oazo-apps/story/9114/ajna-overview-data-pnl

## Description of Changes

Please list the changes introduced by this PR:

- [x] Add PnL method to AjnaPosition class that access cumulative vals as args

## PR Definition of Done

- [x] All checks are passing
- [ ] Code has been self-reviewed and any fixes or improvements noted
- [ ] Documentation has been updated if necessary

### DIFF
--- a/packages/dma-library/src/types/ajna/ajna-position.ts
+++ b/packages/dma-library/src/types/ajna/ajna-position.ts
@@ -137,10 +137,15 @@ export class AjnaPosition implements IAjnaPosition {
     cumulativeWithdrawnUSD: BigNumber,
     cumulativeFeesUSD: BigNumber,
   ): BigNumber {
-    return this.netValue
-      .minus(cumulativeDepositUSD)
-      .plus(cumulativeWithdrawnUSD)
+    if (cumulativeDepositUSD.isZero()) {
+      return ZERO
+    }
+
+    return cumulativeWithdrawnUSD
+      .plus(this.netValue)
       .minus(cumulativeFeesUSD)
+      .minus(cumulativeDepositUSD)
+      .div(cumulativeDepositUSD)
   }
 
   debtAvailable(collateralAmount?: BigNumber) {
--- a/packages/dma-library/src/types/ajna/ajna-position.ts
+++ b/packages/dma-library/src/types/ajna/ajna-position.ts
@@ -34,6 +34,12 @@ export interface IAjnaPosition {
   netValue: BigNumber
   buyingPower: BigNumber
 
+  pnl(
+    cumulativeDepositUSD: BigNumber,
+    cumulativeWithdrawnUSD: BigNumber,
+    cumulativeFeesUSD: BigNumber,
+  ): BigNumber
+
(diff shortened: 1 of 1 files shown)


packages/dma-library/src/types/ajna/ajna-position.ts
```diff
@@ -125,6 +132,17 @@ export class AjnaPosition implements IAjnaPosition {
...
+  pnl(
+    cumulativeDepositUSD: BigNumber,
+    cumulativeWithdrawnUSD: BigNumber,
+    cumulativeFeesUSD: BigNumber,
+  ): BigNumber {
+    return this.netValue
+      .minus(cumulativeDepositUSD)
+      .plus(cumulativeWithdrawnUSD)
+      .minus(cumulativeFeesUSD)
+  }
```
paszkowskiDamian:
```export function calculatePNL(events: VaultEvent[], currentNetValueUSD: BigNumber) {
  const cumulativeDepositUSD = events.reduce(getCumulativeDepositUSD, zero)
  const cumulativeWithdrawnUSD = events.reduce(getCumulativeWithdrawnUSD, zero)
  const cumulativeFeesUSD = events.reduce(getCumulativeFeesUSD, zero)

  if (cumulativeDepositUSD.isZero()) {
    return zero
  }

  return cumulativeWithdrawnUSD
    .plus(currentNetValueUSD)
    .minus(cumulativeFeesUSD)
    .minus(cumulativeDepositUSD)
    .div(cumulativeDepositUSD)
}```

We have slightly different formula in borrow. Can we check which one is correct?

File: packages/dma-library/src/types/ajna/ajna-position.ts [hunk 2]
```diff
@@ -125,6 +132,17 @@ export class AjnaPosition implements IAjnaPosition {
       .minus(this.debtAmount.times(this.quotePrice))
   }
 
+  pnl(
```
piekczyk, line 135: What about taking the same approach as in earn positions? We could potentially calculate it in `getPosition` method (as we do for earn in `getEarnPosition`) and in library only extend interface with 

```
    public pnl: BigNumber,
    public totalEarnings: BigNumber,
```

zerotucks: Yes. I had a chat with Damian about doing that way. But, we thought perhaps keeping the constructor lean and not having to update their usage elsewhere might be simpler.

packages/dma-library/src/types/ajna/ajna-position.ts
zerotucks: My bad. Copied the wrong one like a muppet 🤦‍♂️

paszkowskiDamian: 🫂

paszkowskiDamian (approved):

PR #396, opened 2023-07-19
Open borrow limit index adjust

piekczyk:


packages/dma-library/src/strategies/ajna/borrow/open.ts
```diff
@@ -53,7 +55,7 @@ export const open: AjnaOpenBorrowStrategy = async (args, dependencies) => {
     ethers.utils
       .parseUnits(args.collateralAmount.toString(), args.collateralTokenPrecision)
       .toString(),
-    htp.toString(),
+    limitIndex.toString(),
```
paszkowskiDamian, line 58: I thought we did that change to all strategies already.

paszkowskiDamian:
Ok, I've check that `borrow` already has to fix.

paszkowskiDamian (approved):

PR #399, opened 2023-07-24
Ignore generated ABIs by default

zerotucks:
**CHANGES**
* Generated ABIs are ignored by default
* root `yarn build` now automatically compiles as well
* `lint.yaml` workflow file updated to use consolidated build step that now includes compile


.gitignore
```diff
@@ -12,7 +12,7 @@ tenderly.yaml
 .DS_Store
 
 #Hardhat files
-packages/dma-contracts/abi/generated
+packages/dma-contracts/abis/system/*
```
piotrwitek: @BackToTheCode Is this path correct? For me it looks more like: `packages/abis/system/*`
Following that I would also remove all already commited files in that path as it doesn't make sense to keep them as they are not up to date anyway. What you think?

piotrwitek (approved):
Looks great @BackToTheCode 🎉

PR #441, opened 2023-08-25
feat: check leftover quote token in proxy after repay

halaprix:
## Ticket URL

[sc-11421]

## Description of Changes

Additional checks were added to:
- check if the entire reapy amount has been used to repay debt
- if not - the leftover amount is sent back to the proxy owner
- sine we zero the proxy balance, we gained some gas savings

Affected methods :
- `repayDebt`
- `repayDebtAndWithdrawCollateral`

## How to Test

Please provide instructions on how to test the changes in this PR:

## PR Definition of Done

- [ ] All checks are passing
- [ ] PR is linked to a corresponding ticket
- [ ] PR title is clear and concise
- [ ] Code has been self-reviewed and any fixes or improvements noted (See Code review standards in Notion)
- [ ] Documentation has been updated if necessary

### DIFF
--- a/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
+++ b/packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
@@ -328,9 +328,15 @@ contract AjnaProxyActions is IAjnaProxyActions {
         _pull(debtToken, amount);
         IERC20(debtToken).safeApprove(address(pool), amount);
         (, , , , , uint256 lupIndex_) = poolInfoUtils.poolPricesInfo(address(pool));
+        uint256 balanceBefore = IERC20(debtToken).balanceOf(address(this));
         pool.repayDebt(address(this), amount * pool.quoteTokenScale(), 0, address(this), lupIndex_);
         if (stamploan) {
             pool.stampLoan();
+        }
+        uint256 repaidAmount = balanceBefore - IERC20(debtToken).balanceOf(address(this));
+        uint256 leftoverBalance = amount - repaidAmount;
+        if (leftoverBalance > 0) {
+            _send(debtToken, leftoverBalance);
         }
         emit ProxyActionsOperation("AjnaRepay");
     }
@@ -360,6 +366,7 @@ contract AjnaProxyActions is IAjnaProxyActions {
         _pull(debtToken, debtAmount);
         IERC20(debtToken).safeApprove(address(pool), debtAmount);
         (, , , , , uint256 lupIndex_) = poolInfoUtils.poolPricesInfo(address(pool));
+        uint256 quoteBalanceBefore = IERC20(debtToken).balanceOf(address(this));
         pool.repayDebt(
             address(this),
             debtAmount * pool.quoteTokenScale(),
@@ -368,6 +375,11 @@ contract AjnaProxyActions is IAjnaProxyActions {
             lupIndex_
         );
         _send(collateralToken, collateralAmount);
+        uint256 repaidAmount = quoteBalanceBefore - IERC20(debtToken).balanceOf(address(this));
+        uint256 quoteLeftoverBalance = debtAmount - repaidAmount;
+        if (quoteLeftoverBalance > 0) {
+            _send(debtToken, quoteLeftoverBalance);
+        }
         emit ProxyActionsOperation("AjnaRepayWithdraw");
     }
 
--- a/packages/ajna-contracts/scripts/common/config.ts
+++ b/packages/ajna-contracts/scripts/common/config.ts
@@ -67,5 +67,5 @@ export const CONFIG = {
   initializeStakingRewards: false,
   deployPools: false,
   deployer: "0x8E78CC7089509B568a401f593F64B3074693d25E",
-  logGasUsage: false,
+  logGasUsage: true,
 };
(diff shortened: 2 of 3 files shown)


packages/ajna-contracts/scripts/common/config.ts
```diff
@@ -67,5 +67,5 @@ export const CONFIG = {
   initializeStakingRewards: false,
   deployPools: false,
   deployer: "0x8E78CC7089509B568a401f593F64B3074693d25E",
-  logGasUsage: false,
+  logGasUsage: true,
```
robercano, line 70: Not sure if we want to leave this enabled

packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
```diff
@@ -328,10 +328,16 @@ contract AjnaProxyActions is IAjnaProxyActions {
         _pull(debtToken, amount);
         IERC20(debtToken).safeApprove(address(pool), amount);
         (, , , , , uint256 lupIndex_) = poolInfoUtils.poolPricesInfo(address(pool));
+        uint256 balanceBefore = IERC20(debtToken).balanceOf(address(this));
         pool.repayDebt(address(this), amount * pool.quoteTokenScale(), 0, address(this), lupIndex_);
         if (stamploan) {
             pool.stampLoan();
         }
+        uint256 repaidAmount = balanceBefore - IERC20(debtToken).balanceOf(address(this));
```
robercano, line 336: So there is a prospect problem here:
- Using the balance of the `debtToken` in the contract can be dangerous. ERC777, which is backwards compatible with ERC20, allow the owner of the tokens to be called back when a transfer happens. The owner could then send more debt tokens to the pool contract, modifying the actual balance. I don't think it can be used to drain tokens, but it would definitely make the call fail if the substraction operation is less than 0 (implicit underflow check in Solidity) for line 336
- Also the `leftOverBalance` variable is unsigned, so it cannot be negative, rendering the `if` in line 338 useless

I've checked and the `repayDebt` call from Ajna does not return anything (unfortunately). I think the best approach is to take the balance of debt token BEFORE pulling the tokens in (base balance) and then return anything that exceeds that balance. This preserves the invariant: "`repayDebt` preserves the number of `debtTokens` in the contract"

packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
```diff
@@ -360,6 +366,7 @@ contract AjnaProxyActions is IAjnaProxyActions {
         _pull(debtToken, debtAmount);
         IERC20(debtToken).safeApprove(address(pool), debtAmount);
         (, , , , , uint256 lupIndex_) = poolInfoUtils.poolPricesInfo(address(pool));
+        uint256 quoteBalanceBefore = IERC20(debtToken).balanceOf(address(this));
```
robercano, line 369: We should also talk in the future about these functions, because they seem to have very similar behaviour with some differences. Perhaps we could refactor them to not duplicate code

packages/ajna-contracts/test/pool.dpm.test.ts
```diff
@@ -680,17 +683,84 @@ describe.only("AjnaProxyActions", function () {
...
+        proxy: await wbtc.balanceOf(borrowerProxy.address),
       };
 
       expect(balancesQuoteAfter.borrower).to.be.lt(balancesQuoteBefore.borrower);
       expect(balancesQuoteBefore.borrower).to.be.eq(balancesQuoteAfterDraw.borrower.sub(bn.six.THOUSAND.mul(2)));
       expect(balancesCollateralAfter.borrower).to.be.equal(balancesCollateralBefore.borrower);
+      expect(balancesQuoteAfter.proxy).to.be.eq(ZERO);
+      expect(balancesCollateralAfter.proxy).to.be.eq(ZERO);
     });
+    it.only("should openAndDraw and repayDebtAndWithdrawCollateral", async () => {
```
robercano, line 700: Remove `.only`

packages/ajna-contracts/test/pool.dpm.test.ts
```diff
@@ -680,17 +683,84 @@ describe.only("AjnaProxyActions", function () {
...
+        ajnaProxyActionsContract,
+        poolContract,
+        price,
+        wbtc,
+        borrower,
+        borrowerProxy,
+        bn.six.THOUSAND,
+        bn.eight.TEN
+      );
+      await depositAndDrawDebt(
```
robercano, line 722: The call to `depositAndDrawDebt` is duplicated. Is it on purpose? If so, can you add a comment explaining why?

robercano (changes requested):
I have a major concern on the contract changes. If it is urgent we could deploy with that, but it may be opening the door to some attacks (I'd need more time to figure out if it is possible).

packages/ajna-contracts/test/pool.dpm.test.ts
halaprix: no idea. removing it

packages/ajna-contracts/scripts/common/config.ts
halaprix: true

packages/ajna-contracts/test/pool.dpm.test.ts
halaprix: fixed

packages/ajna-contracts/contracts/ajna/ajna-actions/AjnaProxyActions.sol
robercano: Discussed offline and decided that it is fine. If more tokens are sent to the contract the transaction will revert which is OK, and the if for greater than zero is to prevent zero transfers of the tokens

robercano (approved):

PR #475, opened 2023-09-08
[Tools] Deployment and Operations validation for `OpenDepositBorrow` on L2s

robercano:
## Ticket URL

Please provide a link to the ticket:
https://app.shortcut.com/oazo-apps/story/11762/library-smartcontracts-deploy-opendepositborrow-operation-on-l2s

## Description of Changes

Please list the changes introduced by this PR:

- Added 2 tools for deployment and operations verifications. These tools allowed to verify that the `OpenDepositBorrow` operation in Optimism and Arbitrum have the correct actions deployed and that it only needs updating the `OperationRegistry` definitions

## How to Test

There is a section at the end of the `README.md` explaining how to use the tool and also a couple of Notion documents:
- [Validate Deployment](https://www.notion.so/oazo/Validate-Deployment-Beta-6b35909c77794855a4f0421af7fa7d04)
- [Validate Operations](https://www.notion.so/oazo/Validate-Operations-Beta-c304d5d18e2d41eda41724253e6b92ba)

## PR Definition of Done

- [x] All checks are passing
- [x] Code has been self-reviewed and any fixes or improvements noted (See Code review standards in Notion)
- [x] Documentation has been updated if necessary


jakubswierczek:
There is an issue with building packages: `@oasisdex/dma-contracts: Error: Cannot find module '@typechain/index'`

packages/dma-contracts/tasks/verify-deployment/index.ts
```diff
@@ -0,0 +1,279 @@
...
+const { red, yellow, green } = color
+
+function getValidationStatusString(verificationResult: VerificationResult): string {
+  const verificationPercentage: number = Math.round(
+    (verificationResult.totalValidated / verificationResult.totalEntries) * 100,
+  )
+
+  if (verificationPercentage < 30) {
+    return (
+      red(`${verificationPercentage.toFixed(2)}%`) +
```
jakubswierczek, line 34: It's not obvious what that the `red` function means. I prefer that notation:
```
import { log } from 'console-log-colors';

log.red('red text');
log.green('green text');
log.cyanBright('cyanBright text', 'cyanBright text2');
```

packages/dma-contracts/tasks/verify-deployment/index.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  if (Object.keys(config).length === 0) {
+    return {
+      success: false,
+      totalEntries,
+      totalValidated,
+    }
+  }
+
+  for (const entryName in config) {
```
jakubswierczek: what do you think about:
```
Object.entries(config).forEach(([key, value] ) => { })
```
Then, we won't need to create a `configEntry` and once again iterate through the `config` object.

packages/dma-contracts/tasks/verify-deployment/index.ts
```diff
@@ -0,0 +1,279 @@
...
+    totalEntries++
+
+    const configEntry: ConfigEntry = config[entryName]
+
+    if (isInvalidAddress(configEntry.address)) {
+      console.log(`${entryName}: ❌ (not configured)`)
+      continue
+    }
+
+    totalValidated++
```
jakubswierczek: or even we could do:
```
const validated = Object.entries(config).filter(([key, value] ) => { 
if (isInvalidAddress(configEntry.address)) {
      console.log(`${entryName}: ❌ (not configured)`)
      return false
    }
    return true
})
```

Then we can remove mutating states of the variables.

packages/dma-contracts/tasks/verify-operations/index.ts
```diff
@@ -0,0 +1,232 @@
...
+  for (const operationDefinition of operationDefinitions) {
+    totalEntries++
+
+    let operationHashes: string[]
+    let operationOptionals: boolean[]
+    try {
+      ;[operationHashes, operationOptionals] = await operationRegistry.getOperation(
+        operationDefinition.name,
+      )
+    } catch (e: any) {
```
jakubswierczek: Could we use `unknown` instead of `any`?

jakubswierczek (commented):
I posted some suggestions. We can talk about them later.

packages/dma-contracts/tasks/verify-deployment/index.ts
robercano: The problem here is that I'm returning the colored strings, but not actually logging them, as I do that at the end of the calling function. I could alias those colors to say something like `red_text` or similar. What do you think?

packages/dma-contracts/tasks/verify-deployment/index.ts
robercano: Agree! I wasn't sure what was more readable in TS

packages/dma-contracts/tasks/verify-deployment/index.ts
robercano: If I understand correctly, then your idea is to count the number of validated agains the total number of entries in config, right? If so, I like it better, thanks!

packages/dma-contracts/tasks/verify-operations/index.ts
robercano: Sure! What the idea behind it? Why is `unknown` better?

packages/dma-contracts/tasks/verify-deployment/index.ts
jakubswierczek: oh! okey, it's okey, I didn't catch that :)

packages/dma-contracts/tasks/verify-operations/index.ts
jakubswierczek: Basically it's better for runtime errors. 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type
You can't access any properties without a type assertion.
You can also use syntax to check specific type definitions: 
```
function checkIfItHasMessage(x: unknown): x is { message: string } {
  if (typeof x !== 'object' || x === null) return false
  return 'message' in x
}
```

packages/dma-contracts/tasks/verify-deployment/index.ts
robercano: Added a comment to the declaration of those functions so it clarifies it a bit :)

robercano:
Change the way the `hardhat.config.ts` is used. Now a new file `hardhat.config.base.ts` is used that contains all the config except the tasks imports. This file is used when compiling through `yarn compile`. This avoids the circular dependency with typechain. Then for the rest of the commands, the full `hardhat.config.ts` is used

jakubswierczek (approved):

PR #479, opened 2023-09-13
Base L2 deployment

fenris85:
Base L2 deployment: 

- ServiceRegistry
- DMA core contracts
- DMA Swap
- DPM
- DSProxy


packages/deploy-configurations/configs/base.conf.ts
```diff
@@ -0,0 +1,1211 @@
...
+      },
+      DSGuardFactory: {
+        name: 'DSGuardFactory',
+        deploy: true,
+        address: '0x89010CBE61Ba3A800aFDc20D22E8163c6256E5Ef',
+        serviceRegistryName: SERVICE_REGISTRY_NAMES.common.DS_GUARD_FACTORY,
+        history: [],
+        constructorArgs: []
+      },
+      AccountGuard: {
```
paszkowskiDamian: Do we need those maker contracts?

packages/deploy-configurations/configs/base.conf.ts
```diff
@@ -0,0 +1,1211 @@
...
+    },
+    WLD: {
+      name: 'WLD',
+      address: '0x0000000000000000000000000000000000000000'
+    },
+    WSTETH: {
+      name: 'WSTETH',
+      address: '0x0000000000000000000000000000000000000000',
+      serviceRegistryName: SERVICE_REGISTRY_NAMES.common.WSTETH
+    },
```
paszkowskiDamian: Can we override later those addresses in registry?

packages/deploy-configurations/configs/base.conf.ts
fenris85: Probably we don't need it. But it's deployed already. We could remove it from config for the future (in all networks or on L2s only)

packages/deploy-configurations/configs/base.conf.ts
fenris85: We cannot override in current ServiceRegistry implementation. But this one (or any empty) is not added to SR.

halaprix (approved):

PR #484, opened 2023-09-14
feature: tool to generate operation tuples

robercano:
Tool to generate operation tuple for adding transactions to Safe.

It allows to specify the network that the tuple will be generated for, and it also allows to list all operation names for the given network.

Documentation: https://www.notion.so/oazo/Operations-Tuple-Generation-11c02e5c7de54578a8aa38800492f044


zerotucks (approved):
Wonderful. Tested with `SparkAdjustRiskDown`
Super useful

PR #485, opened 2023-09-15
:bug: Flashloans adjustments for Aave Like Protocols

jakubswierczek:
## Ticket URL

[Shortcut](https://app.shortcut.com/oazo-apps/story/11917/frontend-library-issues-with-flashloans)

## Description of Changes
- Return flash loan information with `Multiply Strategy`
- Possibility to force flash loan token for Aave & Spark (Although that will work only on L2s)
- Refactor the Close operation. After depositing a flash loan, we were taking the MAX_UNIT amount of collateral. In the case of short positions (for example, with DAI on Mainnet), that would break the transaction. I have changed the behavior, and now, when the collateral and flash loan assets are the same, I withdraw only the necessary amount for the swap. 

## How to Test

Please try to play around with Aave v3 and Spark on the [preview deployment](https://kh8439-3000.csb.app/). In the order information you will see the flashloan amount. 

## PR Definition of Done

- [x] All checks are passing
- [x] Code has been self-reviewed and any fixes or improvements noted (See Code review standards in Notion)
- [x] Documentation has been updated if necessary


packages/domain/src/flashloans.ts
```diff
@@ -39,6 +39,7 @@ export const transientCollateralFlashloan: TransientCollateralFlashloan = (
   )
     .times(oraclePrice)
     .div(maxLoanToValueWhenCollateralising.times(ONE.minus(FLASHLOAN_SAFETY_MARGIN)))
+    .abs()
```
robercano, line 42: Why is this `.abs()` needed?

packages/dma-library/src/strategies/aave/common/get-flashloan-token.ts
```diff
@@ -1,15 +1,32 @@
...
 }
 
 export function getFlashloanToken({
   network,
   addresses,
-}: FlashloanDependencies): WithFlashloanToken {
+  protocol,
+  debt,
+}: FlashloanDependencies): WithFlashLoanArgs {
+  if (protocol === 'Spark') {
```
robercano, line 22: Is it possible to use an `enum` here instead of a harcoded string?

packages/dma-library/src/strategies/aave-like/multiply/common/build-flashloan-sim-args.ts
```diff
@@ -1,23 +1,29 @@
...
+import { IPositionTransitionParams } from '@domain'
 import BigNumber from 'bignumber.js'
 
 /**
  * TODO: Remove the need for this function
  * Flashloan calculations should not be part of the adjustToTargetRiskRatio logic
  */
-export function buildFlashloanSimArgs(
-  flashloanTokenAddress: string,
+export async function buildFlashloanSimArgs(
```
robercano: Why is this `async`? I don't see any calls that need `await`

packages/dma-library/src/strategies/aave-like/multiply/close/types.ts
```diff
@@ -1,16 +1,28 @@
...
 
 export type AaveLikeExpandedCloseArgs = AaveLikeCloseArgs &
   StrategyParams.WithProtocolData & {
     collateralToken: AaveLikeCloseArgs['collateralToken'] & { address: string }
     debtToken: AaveLikeCloseArgs['debtToken'] & { address: string }
-    flashloanToken: WithFlashloanToken['flashloanToken'] & { address: string }
+  } & WithFlashLoanArgs
+
+export type CloseFlashloanArgs = {
+  token: {
```
robercano, line 18: I think we could create a `Token` type being:
```
export type Token : {
  address: string
  precision: number
  symbol: string
}
```

and then a `TokenAmount` type being:
```
export type TokenAmount:  {
  token: Token
  amount: BigNumber
}
```

To try to start having a bit more semantics in the code

robercano (changes requested):
Looks good in general! I'm missing the related tests, though. Is this done somewhere else?

packages/dma-library/src/strategies/aave/common/get-flashloan-token.ts
jakubswierczek: The protocol variable is of type: `export type AaveLikeProtocol = 'AAVE' | 'AAVE_V3' | 'Spark'`, so it's not hardcoded per se. ;)

packages/dma-library/src/strategies/aave-like/multiply/common/build-flashloan-sim-args.ts
jakubswierczek: My apologies, I forgot to remove it after conducting my tests.

packages/dma-library/src/strategies/aave-like/multiply/close/types.ts
jakubswierczek: Yes it's a good idea!

robercano (approved):
Agreed to make the following changes in a future PR instead of in this one:
- Use `Protocol.SPARK` from `dma-library/src/types` instead of hardcoded string
- Review the use of `.abs()` when calculating flashloans to see if it is still needed
- Agreed to use `Token` and `TokenAmount` types

PR #501, opened 2023-10-23
feat: single entry addition for serv.reg. tool

robercano:
Small addition for the Service Registry tool to support single entries as parameter. Now you can do:

```bash
yarn service:pushconfig common.OneInchAggregator --network base
```

And it will push that single entry to the service registry.

Also added One Inch aggregator address to Base service registry


packages/dma-contracts/tasks/service-registry/index.ts
```diff
@@ -104,59 +166,16 @@ async function pushConfigToRemote(configPath: string, network: string, ethers: a
...
+  } else {
+    // Multiple entries
+    for (const serviceName of Object.keys(property)) {
+      const service = property[serviceName] as ConfigEntry
 
-    if (!service.serviceRegistryName) {
-      console.log(`[✅][LOCAL CONFIG] ${service.name}: ${service.address}`)
-      continue
+      await pushEntryToRemote(serviceRegistry, service)
     }
```
halaprix, line 178: cant comment below, but maybe this could be refactored to subtasks ?

```ts
   if (taskArgs.pushconfig) {
      await pushConfigToRemote(taskArgs.pushconfig, network, ethers)
    } else if (taskArgs.showlocal) {
      showLocalConfig(taskArgs.showlocal, network)
    } else if (taskArgs.showremote) {
      await showRemoteConfig(taskArgs.showremote, network, ethers)
    } else {
      throw new Error('Either --pushconfig, --showlocal or --showremote must be specified')
    }
```

packages/dma-contracts/tasks/service-registry/index.ts
```diff
@@ -48,37 +77,70 @@ async function showRemoteConfig(configPath: string, network: string, ethers: any
...
+  console.log('\n==============================\n')
+}
+
+async function pushEntryToRemote(serviceRegistry: ServiceRegistry, service: ConfigEntry) {
+  if (!service.serviceRegistryName) {
+    console.log(`[✅][LOCAL CONFIG] ${service.name}: ${service.address}`)
+    return
+  }
+
+  let serviceAddress
```
halaprix:
```suggestion
    let serviceAddress: undefined | string
```

halaprix (changes requested):

halaprix (approved):

packages/dma-contracts/tasks/service-registry/index.ts
robercano: Discussed offline: subtasks from Hardhat don't offer much functionality in this case. It would be great to have subcommands in a main command so each subtask takes care of each subcommand, but in the case of Hardhat's subtasks it just separates the help message, but still the main command has to handle the parameter definition (or at least that's what is inferred from the Hardhat examples). I'll leave it like that and I'll try to figure out if there's something else that can be done in that direction

PR #521, opened 2023-11-21
feat: add refinance framework + AAVE-->AAVE refinance

robercano:
## Ticket URL

[SC-12920](https://app.shortcut.com/oazo-apps/story/12920/sc-create-aave-aave-refinance-operation)

## Description of Changes
- Added Refinance framework to easily define new refinance operations. Added a [README](https://github.com/OasisDEX/oasis-earn-sc/blob/023288c9cdb4d5f3ab7d3f285a5e487f8685bedc/packages/dma-library/src/operations/refinance/README.md) explaining how to use the framework and how to expand it with new protocols
- Added AAVE-->AAVE refinance operation
- Added tests for AAVE-->AAVE refinance
- Added helpers to interact with `MockExchange` for fake swaps
- Moved 'Protocol' definition from `dma-library` to `dma-common` so it can be used from `dma-contracts` without creating a cyclic dependency
- Added a tool to retrieve the available Refinance operations and the operations definitions. This is in preparation for a tool that allows to add the refinance operations to the `OperationsRegistry`

## How to Test

Preparation:
```bash
$ yarn
$ cd packages/dma-contracts
$ yarn compile
```

Then either:
```bash
$ yarn  test
```
to run all the tests, or:
```bash
$ yarn test:e2e
```
to run the E2E tests that include refinance as well

## PR Definition of Done

Please ensure the following requirements have been met before marking the PR as ready for review:

- [✔️] All checks are passing
- [✔️] PR is linked to a corresponding ticket
- [✔️] PR title is clear and concise
- [✔️] Code has been self-reviewed and any fixes or improvements noted (See Code review standards in Notion)
- [✔️] Documentation has been updated if necessary

PR #525, opened 2023-11-24
[DMA V2] Gas-Optimised

fenris85:
Gas optimisations changes cherry picked from old Georgi's branch (https://github.com/OasisDEX/oasis-earn-sc/tree/gg/v2-sc-gas-optimized) 

Main changes regard OperationRegistry and how we are storing operations - now it's hash of actions hashes along with operation name. 

OperationExecutor has additional method for Automation purposes to get opName by hash easily.

Actions that were using read or write to the storage has changed (Aave V3 and Common so far).

Update to Actions required by Migration feature (add onBehalf to Aave V3 Payback, added params mapping Aave V3 Borrow,  add Common TokenBalance Action)


packages/dma-contracts/contracts/core/OperationsRegistry.sol
```diff
@@ -29,41 +29,45 @@ contract OperationsRegistry {
...
+   * that are used in the specific operation and hash them using keccak256.
+   * The order of the actions matters since different order will result in different hash value
+   * hence difference operation.
+   * @param name Name of the operation that is going to be stored
+   * @param operationHash A hash of all action hashes ( concatenated )
    */
-  function addOperation(StoredOperation calldata operation) external onlyOwner {
-    operations[operation.name] = operation;
+  function addOperation(string memory name, bytes32 operationHash) external onlyOwner {
+    operations[operationHash] = bytes32(bytes(name));
```
halaprix, line 52: I'd add the check if the `operations[operationHash] != bytes32("")`

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -1,120 +1,131 @@
...
+  function executeOp(Call[] memory calls) public payable {
+    StorageSlot.TransactionStorage storage txStorage = StorageSlot.getTransactionStorage();
+    delete txStorage.actions;
+    delete txStorage.returnedValues;
 
-    opStorage.clearStorage();
-    (bytes32[] memory actions, bool[] memory optional) = opRegistry.getOperation(operationName);
-    opStorage.setOperationActions(actions, optional);
     aggregate(calls);
+    bytes32 operationName = this.getOperation(keccak256(abi.encodePacked(txStorage.actions)));
```
halaprix: is `.this` necessary ?

halaprix (changes requested):

packages/dma-contracts/contracts/core/OperationExecutor.sol
halaprix: that won't work I think, if the `executeOp` is delegatecall'ed - as `this` will be the proxy

packages/dma-contracts/contracts/actions/aave/v3/Borrow.sol
```diff
@@ -15,26 +15,36 @@ import { IPoolV3 } from "../../../interfaces/aaveV3/IPoolV3.sol";
...
  * @notice Borrows token from AAVE's lending pool
  */
-contract AaveV3Borrow is Executable, UseStore {
-  using Write for OperationStorage;
+contract AaveV3Borrow is Executable, UseStorageSlot {
+  using Read for StorageSlot.TransactionStorage;
+  using Write for StorageSlot.TransactionStorage;
 
-  constructor(address _registry) UseStore(_registry) {}
+  ServiceRegistry internal immutable registry;
```
robercano: I'd suggest to abstract this bit away in a separate contract:
```
contract UseRegistry {
  ServiceRegistry private immutable _registry;
  
  constructor(ServiceRegistry registry_) {
    _registry = registry_
  }
  
 function getRegisteredService(string service) return address
 {
   return _registry.getRegisteredService(service);
 }
}
```

Right now the end result is very similar, but moving forward this abstracts access to the registry and gives us more flexibility to introduce checks and advanced functionality if needed, without having to change all the actions

packages/dma-contracts/contracts/actions/common/PullToken.sol
```diff
@@ -20,7 +20,12 @@ contract PullToken is Executable {
   function execute(bytes calldata data, uint8[] memory) external payable override {
     PullTokenData memory pull = parseInputs(data);
 
-    IERC20(pull.asset).safeTransferFrom(pull.from, address(this), pull.amount);
+    IERC20 token = IERC20(pull.asset);
+
+    if (pull.amount == type(uint256).max) {
+      pull.amount = token.balanceOf(pull.from);
```
robercano, line 26: This is a tricky one: do we want to pull the whole balance or rather the whole allowance? We will never be able to pull more than the allowance, but if the allowance is also `type(uint256).max` then we can pull the balance. Should we use the smaller of the two?

packages/dma-contracts/contracts/actions/common/PullToken.sol
```diff
@@ -20,7 +20,12 @@ contract PullToken is Executable {
...
+    }
+    token.transferFrom(pull.from, address(this), pull.amount);
```
robercano: This will fail for tokens like USDT, please use `safeTransferFrom` instead. Also, if possible, use `SafeERC20` from OpenZeppelin library, instead of our custom implementation of `SafeERC20`. If the size of the contract is a concern, then we can copy/paste the latest version, but I'd suggest just using it

packages/dma-contracts/contracts/core/OperationExecutor.sol
robercano, line 1: If we are going to touch the executor I'd suggest moving errors and events into an interface file to clean it up a bit. The interface would contain the public functions, errors, enums and events. ALso other interfaces (like IProxy) if it makes sense.

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -1,120 +1,131 @@
-// SPDX-License-Identifier: AGPL-3.0-or-later
+//SPDX-License-Identifier: Unlicense
```
robercano: Is this correct? All of our contracts are AGPL

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -1,120 +1,131 @@
...
+   * @param calls List of action calls to be executed.
    */
-  function executeOp(Call[] memory calls, string calldata operationName) public payable {
-    OperationStorage opStorage = OperationStorage(registry.getRegisteredService(OPERATION_STORAGE));
-    opStorage.lock();
-    OperationsRegistry opRegistry = OperationsRegistry(
-      registry.getRegisteredService(OPERATIONS_REGISTRY)
-    );
+  function executeOp(Call[] memory calls) public payable {
+    StorageSlot.TransactionStorage storage txStorage = StorageSlot.getTransactionStorage();
```
robercano, line 88: I think `OperationExecutor` should inherit from `UseStorageSlot` and we should add a `library Clear` to the `UseStorageSlot.sol` that does this:

```
library Clear {
  function clear(StorageSlot.TransactionStorage storage _storage)
  {
    delete _storage.actions;
    delete _storage.returnedValues;
  }
}
```

It abstracts what clearing the storage means which is not a concern that `OperationExecutor` should have

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -1,120 +1,131 @@
...
    **/
   event Operation(bytes32 indexed name, Call[] calls);
 
-  constructor(ServiceRegistry _registry) {
-    registry = _registry;
+  constructor(
+    ServiceRegistry _registry,
+    OperationsRegistry _operationsRegistry,
+    ChainLogView _chainLogView,
+    address _balanacerVault
```
robercano:
```suggestion
    address _balancerVault
```

packages/dma-contracts/contracts/core/OperationExecutor.sol
robercano: Also doing that forces the call to be external, which is more costly

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -126,83 +137,111 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
     uint256 fee,
     bytes calldata data
   ) external override returns (bytes32) {
+    checkIfFlashloanIsInProgress();
```
robercano, line 149: We should use the `ReentrancyGuard` from OpenZeppelin here. It does exactly the same thing and it has been battle-tested for quite a bit now

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -126,83 +137,111 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
...
+    uint256 paybackAmount = amount + fee;
+    uint256 funds = IERC20(asset).balanceOf(address(this));
+    if (funds < paybackAmount) {
+      revert InsufficientFunds(funds, paybackAmount);
+    }
+
+    IERC20(asset).safeApprove(mcdFlash, paybackAmount);
 
-    IERC20(asset).safeApprove(lender, paybackAmount);
     return keccak256("ERC3156FlashBorrower.onFlashLoan");
```
robercano: We could cache the value of this `keccak256` in the constructor as immutable and then return it here, saving some gas

packages/dma-contracts/contracts/core/OperationExecutor.sol
```diff
@@ -126,83 +137,111 @@ contract OperationExecutor is IERC3156FlashBorrower, IFlashLoanRecipient {
...
+   * @param data Any calldata sent to the contract for execution later in the callback
+   */
   function receiveFlashLoan(
     IERC20[] memory tokens,
     uint256[] memory amounts,
     uint256[] memory feeAmounts,
     bytes memory data
   ) external override {
+
+    checkIfFlashloanIsInProgress();
```
robercano, line 186: Same comment for `ReentrancyGuard`

packages/dma-contracts/contracts/core/OperationsRegistry.sol
```diff
@@ -29,41 +29,45 @@ contract OperationsRegistry {
...
+   * To get the operationHash, one should concatenated all hashes of the actions
+   * that are used in the specific operation and hash them using keccak256.
+   * The order of the actions matters since different order will result in different hash value
+   * hence difference operation.
+   * @param name Name of the operation that is going to be stored
+   * @param operationHash A hash of all action hashes ( concatenated )
    */
-  function addOperation(StoredOperation calldata operation) external onlyOwner {
-    operations[operation.name] = operation;
+  function addOperation(string memory name, bytes32 operationHash) external onlyOwner {
```
robercano, line 48: This is breaking the public interface and at least 2 files are affected:
- `operations-registry.ts` which I believe is used in tests and deployment system
- `tasks/operation-registry` which is one of the tools I did.

Fix should be trivial though

packages/dma-contracts/contracts/core/OperationsRegistry.sol
```diff
@@ -29,41 +29,45 @@ contract OperationsRegistry {
...
-   * @return actions Returns an array of Actions and array for optionality of coresponding Actions
+   * @notice Returns either a valid name of an added operation or reverts with an error if no operation is found.
+   * @param operationHash A hashed value of all action hashes ( concatenated ).
    */
-  function getOperation(
-    string memory name
-  ) external view returns (bytes32[] memory actions, bool[] memory optional) {
-    if (keccak256(bytes(operations[name].name)) == keccak256(bytes(""))) {
-      revert("Operation doesn't exist");
+  function getOperationName(bytes32 operationHash) external view returns (bytes32 name) {
```
robercano, line 62: It should return `(string name)` so it uses the same interface as `addOperation`. Otherwise I'd change `addOperation` to use `bytes32 name` as well

packages/dma-contracts/contracts/core/OperationsRegistry.sol
```diff
@@ -29,41 +29,45 @@ contract OperationsRegistry {
...
-    actions = operations[name].actions;
-    optional = operations[name].optional;
   }
-}
+
+  /**
+   * @notice Changes the owner of the contract
+   * @param newOwner The address of the new owner of the Operations Registry
+   */
+  function transferOwnership(address newOwner) public onlyOwner {
```
robercano: This should be external only

robercano (changes requested):
Some of the changes will break this [PR](https://github.com/OasisDEX/oasis-earn-sc/pull/521), so we need to merge that one before we merge this.

packages/dma-contracts/contracts/actions/common/PullToken.sol
fenris85: One of the use cases here is that frontend would set up allowance before the tx. And thus exact amount is not known before the tx (it could be aToken for example) the allowance is slightly bigger to cover any additional amount. But if we would pull this allowed amount it might turn out there is little less than the allowance. So I would stick to the token balance, as it's specific to use cases we have right now.

robercano: Thanks for the insight!

PR #705, opened 2024-10-16
Extend Swap Contract with a new behavior to take a fixed fee

piotrwitek:
This pull request introduces a new `FeeType` enum to the Swap contract, introducing ability to take a fee in the fixed amount alongside existing percentage fee.

Additionally implemented couple of chores, like enhanced error handling by providing clear messages for missing environment variables. the Hardhat Tenderly setup, and various tests have been improved to ensure better reliability. Documentation has been added to support these changes.


packages/dma-contracts/contracts/swap/Swap.sol
```diff
@@ -127,13 +128,41 @@ contract Swap {
...
   }
 
+  function _collectFixedFee(
+    address asset,
+    uint256 fromAmount,
+    uint256 fee
+  ) internal returns (uint256 amount) {
+    amount = fromAmount.sub(fee);
+
+    if (fee > 0) {
```
piotrwitek, line 139: @zerotucks should I also add a check somewhere here to guard when the fee is greater than fromAmount?

piotrwitek: Done, if fee is greater than fromAmount then we'll take whatever is available as fee. Confirmed by Chris.

robercano (approved):
Looks great!! I've mainly reviewed the solidity changes and the look awesome

halaprix (approved):

